PR #7, opened 2025-01-22
Multi Quorum ECDSAStakeRegistry

hudsonhrh:
Created new ECDSAStakeRegistry, IECDSAStakeRegistryEventsAndErrors, and ECDSAStakeRegistryStorage contracts in src that are now referenced by the layer middleware deployment 

Quorum was changed to an array that holds multiple strategies in the storage 

The stake registry now has an addQuorum function for adding new quorums. Also added to initializer.

The updateQuroumConfig function was modified to take in the index of what quorum you would like to select. 

So now the flow is add quorum then update the config to select that index


rubydusa:
We've switched the `eigenlayer-middleware` release we're working from in 4985e2a, is it possible to modify these changes to use the new Eigenlayer's `ECDSAStakeRegistry`?

Also in order to not have to do this the next time we update the dependency version maybe it's better to inherit Eigen's `ECDSAStakeRegistry` and make the changes in a child contract

RonTuretzky:
Closed https://github.com/BreadchainCoop/wavs-el-env/issues/10

Issue #10, opened 2025-02-04
Set the metadata url


secbajor:
- [ ] Metadata url  - utility that reads from avs_deploy and sets the metadata url 
     - [ ]   if on testnet deployment mode, add checking that this file exists 
     - [ ] set the metadata url on the ServiceManager

RonTuretzky:
Closed with #15

PR #15, opened 2025-02-10
Metadata URL Updated

hudsonhrh:
Edited LayerMiddlewareDeployment Lib to write the metadata url to the deploy json 

Closes issue #10


rubydusa (approved):

RonTuretzky (changes requested):
Storing the metadata url in the 17000.json is not a good pattern because this is a file created with each deployment, and will get overwritten. Please change to read be expected in an ENV variable

hudsonhrh:
I just also added it there for reference. If you look at the deployContracts function in the layer middleware deployment lib I added the url there hardcoded so that the avs_deploy.json and that 1700.json do contain the url. I could make it so instead the url is passed into that function and then update the deployment script to pass in the url and have it read it from the deployments/core/1700.json or an ENV

RonTuretzky:
> I just also added it there for reference. If you look at the deployContracts function in the layer middleware deployment lib I added the url there hardcoded so that the avs_deploy.json and that 1700.json do contain the url. I could make it so instead the url is passed into that function and then update the deployment script to pass in the url and have it read it from the deployments/core/1700.json or an ENV

All good , we should probably add that json to the gitignore at some point nearing mainnet release 

Cc: @secbajor

PR #23, opened 2025-02-16
22 registration env is based on testnet rpc url instead of deploy env

Astodialo:


RonTuretzky:
closed because these utils will be deprecated

PR #45, opened 2025-03-27
Solidity port

RonTuretzky:
As defined by @dakom @ethanfrey this pr 

- [x] Introduces the interfaces for all necessary sol code 
- [x] refactors Layer > Wavs for consistency with WAVS naming


dakom (approved):
utACK (just learned that term recently :P)

PR #48, opened 2025-03-27
Adding envelope encoding

### LINKED ISSUE #47: Define `Envelope` in IWavsServiceHandler interface
We pull in the `Envelope` from this interface via alloy in WAVS [here](https://github.com/Lay3rLabs/WAVS/blob/a8da4d8b25ca846e86ecd1272d72d574b7f377a8/packages/types/src/solidity_types.rs#L13) and [here](https://github.com/Lay3rLabs/WAVS/blob/a8da4d8b25ca846e86ecd1272d72d574b7f377a8/packages/types/src/packet.rs#L1)

We then [sign this envelope](https://github.com/Lay3rLabs/WAVS/blob/a8da4d8b25ca846e86ecd1272d72d574b7f377a8/packages/utils/src/eth_client/signing.rs#L16) and [submit it on-chain](https://github.com/Lay3rLabs/WAVS/blob/a8da4d8b25ca846e86ecd1272d72d574b7f377a8/packages/utils/src/eth_client/signing.rs#L55)

Since the `Envelope` is pulled in from this repo, it needs to be defined in the `IWavsServiceHandler` interface

RonTuretzky:


RonTuretzky:
@dakom @ethanfrey is this good to merge?

contracts/interfaces/IWavsServiceHandler.sol
```diff
@@ -2,9 +2,19 @@
 pragma solidity ^0.8.0;
 
 interface IWavsServiceHandler {
+    struct SignatureData {
```
JakeHartnell, line 5: Can we document these things? : )

What is a `referenceBlock` for example?

dakom:
@RonTuretzky I just spoke to @ethanfrey ,  let's make one more change: `eventId` should be an address type, not uint256

RonTuretzky:
> @RonTuretzky I just spoke to @ethanfrey ,  let's make one more change: `eventId` should be an address type, not uint256

Curious , why are we using an address type and not bytes?

dakom:
`bytes20`?

ethanfrey:
Just want to make it bytes20, which is aliased to address, right?

Point is, 20 bytes is enough for collision resistance for ethereum accounts, so should be plenty for our eventuuids, and then we have 12 bytes to play with for other non-opaque routing info (like a sequence number)

(This was in my diagrams of the event envelope. I guess shared photos on drive are the best way to document issues)

RonTuretzky:
> Just want to make it bytes20, which is aliased to address, right?
> 
> Point is, 20 bytes is enough for collision resistance for ethereum accounts, so should be plenty for our eventuuids, and then we have 12 bytes to play with for other non-opaque routing info (like a sequence number)
> 
> (This was in my diagrams of the event envelope. I guess shared photos on drive are the best way to document issues)

Ok understood, will follow up

RonTuretzky:
> Just want to make it bytes20, which is aliased to address, right?
> 
> Point is, 20 bytes is enough for collision resistance for ethereum accounts, so should be plenty for our eventuuids, and then we have 12 bytes to play with for other non-opaque routing info (like a sequence number)
> 
> (This was in my diagrams of the event envelope. I guess shared photos on drive are the best way to document issues)

Me @hudsonhrh and @rubydusa all agree it's better to use bytes20

contracts/interfaces/IWavsServiceHandler.sol
```diff
@@ -2,9 +2,19 @@
...
+        address[] operators;
```
ethanfrey, line 6: Nit: if we have recover address from signature, why do we need to pass the addresses in here?

ethanfrey: (I mean this is probably a question about ServiceManager or StakeRegistry contract, just exposed here)

contracts/interfaces/IWavsServiceHandler.sol
```diff
@@ -2,9 +2,19 @@
...
+        bytes[] signatures;
+        uint32 referenceBlock;
+    }
+    struct Envelope {
+        bytes payload;
+        uint256 eventId;
```
ethanfrey: Let's make this `bytes20`.
For padding, I assume better to put it in front, and give another `bytes12` fields there, which can be called "unused" or "ordering" if you really want to start prepping for that feature

contracts/src/WavsServiceManager.sol
```diff
@@ -161,17 +162,17 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
...
         bytes32 ethSignedMessageHash = ECDSAUpgradeable.toEthSignedMessageHash(message);
         bytes4 magicValue = IERC1271Upgradeable.isValidSignature.selector;
-
+        bytes memory signatureDataBytes = abi.encode(signatureData.operators, signatureData.signatures, signatureData.referenceBlock);
         // If the registry returns the magicValue, signature is considered valid
         if( magicValue !=
             ECDSAStakeRegistry(stakeRegistry).isValidSignature(
                 ethSignedMessageHash,
-                signature
+                signatureDataBytes
```
ethanfrey, line 175: Yup, this is taken from ECDSAStakeRegistry

ethanfrey (commented):
Good stuff. Some minor changes and let's get this merged

contracts/interfaces/IWavsServiceHandler.sol
RonTuretzky: does [this](https://github.com/BreadchainCoop/eigenlayer-middleware/blob/848032a7e22b2e6a2ab7835a45e9b39bb16d41a5/src/unaudited/ECDSAStakeRegistry.sol#L417) help explain it?

contracts/interfaces/IWavsServiceHandler.sol
RonTuretzky: yea it's to grab the stake, the logic is [here](https://github.com/BreadchainCoop/eigenlayer-middleware/blob/848032a7e22b2e6a2ab7835a45e9b39bb16d41a5/src/unaudited/ECDSAStakeRegistry.sol#L419)

ethanfrey (approved):
I changed eventId to bytes20. Now good to merge from my side.
@dakom are we good here for WAVS?

dakom:
@ethanfrey thanks! LGTM. This will close https://github.com/Lay3rLabs/wavs-middleware/issues/47

Issue #57, opened 2025-03-31
Final Trim


RonTuretzky:
Remove any remanants from the `hello-world-avs` repo

ethanfrey:
All remnants are now in the Rust code, which will be tackled in #42 and #52 

Closing this one

PR #62, opened 2025-04-02
Trim out solidity, better metadata uri flow

### LINKED ISSUE #36: Add better config to allow to set metadata_uri
When running `docker/deploy.sh` it parses the contracts/deployment/core/17000.json to get the metadataURI.

This should be configurable and set when calling deploy.

### LINKED ISSUE #38: Clean up deploy script flow
## Pass through deployed contracts

We print out the deployed service manager address, but this should be written to a file somewhere on the host filesystem to pass to future scripts.
This requires mounting some location to the docker image.

## Clean up set_service_uri.sh

Calling the script is a bit ugly, needs entrypoint and is a bit finicky.
Let's make it more robust.

* Use the mounted dir to get the service manager
* Make passing in metadata_uri cleaner for the cli script

## Better deploy.sh env settings

We only pass in a few env vars and there were definitely more in wavs-el-env repo that we should pull into `env.example`
Let's add them and figure out a better way to pass a lot in via docker run

ethanfrey:
Closes #38 
Closes #36 
Part of #57

This removes and trims down all the solidity files not used anymore. It also removes using some weird hardcoded metadata uri from core deployment, and allows the project to set it in their .env file.


dakom (approved):
My review is just reading through the code and looking out for any glaring bugs, and it all looked fine to me. Can't speak to the bigger picture like if it truly removed all remnants of hello world AVS or manages all the addresses we expect, someone more familiar with Eigenlayer would need to review from that perspective if needed

ethanfrey:
Thanks for the review

I pinged the breadchain guys. I’d be happy for a review from someone before merging, but will base any future work off of this PR

Issue #61, opened 2025-04-02
Clean up ERC20 usage in LayerMiddlewareDeployer


ethanfrey:
Looking at the code, I can see that `ERC20Mock.sol` is actually used in the main deployment script `LayerMiddlewareDeployer.s.sol`. It's used to:

* Create a mock ERC20 token (token = new ERC20Mock())
* Deploy a new strategy using that token (helloWorldStrategy = IStrategy(StrategyFactory(coreDeployment.strategyFactory).deployNewStrategy(token)))
* Set up quorum parameters with that strategy

We need to keep `test/ERC20Mock.sol` since it's used as part of the deployment process to create a test token for the strategy. The file being in the test directory is a bit misleading since it's actually used in production deployment code.

However, this smells like some legacy test code and not appropriate for testnet/mainnet deployments. Don't we want to reuse an existing LST? Or if we do want to deploy our own ERC20, shouldn't we use a production quality contract (likely from lib/open-zeppelin or such) rather than a Mock?

I'd love someone from breadchain to investigate and report on the proper strategy here

ethanfrey:
In particular, the usage here seems suspect for any real deployment:

```solidity
        token = new ERC20Mock();
        helloWorldStrategy = IStrategy(StrategyFactory(coreDeployment.strategyFactory).deployNewStrategy(token));

        quorum.strategies.push(
            StrategyParams({strategy: helloWorldStrategy, multiplier: 10_000})
        );
```

Can we accept a real strategy manager as an argument? Maybe we can provide a default for the most standard one on testnet. Is there a list of the best strategies to use on Eigenlayer? Can you point to some Eigenlayer docs on this?

RonTuretzky:
so we actually already do this at the operator set level, not at the middleware level [here](https://github.com/Lay3rLabs/wavs-middleware/blob/b49a5618885bb84810c80ab2a0136c66ba472dcc/docker/deploy.sh#L69) 

we can update the quorum configuration to match , it should work as the operators are being funded with the correct LST in this [step](https://github.com/Lay3rLabs/wavs-middleware/blob/b49a5618885bb84810c80ab2a0136c66ba472dcc/docker/register.sh#L62)

RonTuretzky:
> In particular, the usage here seems suspect for any real deployment:
> 
>         token = new ERC20Mock();
>         helloWorldStrategy = IStrategy(StrategyFactory(coreDeployment.strategyFactory).deployNewStrategy(token));
[...]

PR #65, opened 2025-04-04
Update register to accept private key as argument

### LINKED ISSUE #60: [register] Accept AVS Private Key from WAVS node
When we register, it currently [generates a new random key](https://github.com/Lay3rLabs/wavs-middleware/blob/dev/docker/register.sh#L57-L59) to register with. However, for our workflow, this must be a service-specific private key generated by the WAVS node once the service was added to it.

register.sh script should be updated such that it takes the private key via a CLI argument. (We don't want it in the env file and it seems problematic to combine env file and individual env variables, so let's just make it a CLI arg).

@ueco-jb please check the [general flow for adding a wavs node in the template](https://github.com/Lay3rLabs/wavs-foundry-template/pull/105/files#diff-3ac67425b1a52c093032d3772e4bf641c0e5fa828ab4efd90d311432f7efc550) and add any more notes here. The last step to register the operator was a TODO and needs to be fully defined so the wavs-middleware script provides what we need

ethanfrey:
Closes #60 

Merge this after #64 is approved and merged

This takes a private AVS Key from the WAVS node to be used for registration.

Also added a lot of comments on register.sh that should be fixed up in follow-up issues.
In particular, https://github.com/Lay3rLabs/wavs-middleware/issues/61 and using a real LST on the dry run as well so we can test the same config we are really going to deploy with.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -51,5 +51,8 @@ docker run --rm --network host --env-file .env  -v ./.nodes:/root/.nodes   --ent
 Register: 
 
 ```bash
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware
+# TODO: get the private AVS key (0x...) for this service from the WAVS node
+AVS_KEY=0x974b676703542ff93841c3daeeabcbfdb6ba62101856e22d5fb6b9d2f9db42fd
+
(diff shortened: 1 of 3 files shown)


ethanfrey:
I'd love a review by @RonTuretzky but don't want to be blocked

Issue #32, opened 2025-03-20
Move `@wavs/solidity` npm package here


dakom:
Should include everything from https://github.com/Lay3rLabs/WAVS/tree/main/contracts/solidity so that WAVS merely needs to import the required interfaces

ethanfrey:
There are some open questions on how to distribute these contracts best. Possibilities seem to be:

* npm package
* git submodules
* cargo package (with alloy bindings)?

We should come to agreement on the way to publish before packaging stuff. I would lean toward whatever helps the end user the most, and defer to @Reecepbcups for info on what the users want. We can figure how to use stuff internally even if it isn't ideal, but want to prioritize making onboarding easy.

reecepbcups:
npm package is the most ideal but not required. Can hack gitsubmodule like functionality in package.json upstream like we have to do for forge. Just a depth of 1 clone

```
  "dependencies": {
    "forge-std": "github:foundry-rs/forge-std#v1.9.6",
    "@wavs/solidity": "^0.3.0",
  }
```

So long as I can do something like either of these we're good. 🚀

ethanfrey:
Question: @Reecepbcups you use the package.json there for building/testing the ServiceHandler contracts? Or for frontend code?

Just trying to figure out best practices for package management and where it is consumed. Cuz I insist using git submodules is only a hack for languages without a proper package manager (looking at you C) and modern languages usually have a better, though language-specific, method. So knowing what tools are consuming it is helpful to shape how to distribute

dakom:
fwiw foundry doesn't _really_ want submodules either, they just don't know what they want yet (https://github.com/foundry-rs/foundry/issues/2618)

reecepbcups:
I use it for building/testing the ServiceHandler contracts. I don't do any front-end work. I assume it would be the same for that though.

ethanfrey:
Building testing the ServiceHandler via forge? Or which tool?

Could you link the lines here? The point here is to make life as easy as possible for the template so I’d love to review the whole workflow from perspective of the template

reecepbcups:
https://github.com/Lay3rLabs/wavs-foundry-template/blob/main/src/contracts/WavsSubmit.sol#L4-L8

where `"@wavs/interfaces/IWavsServiceHandler.sol";` is `@wavs/=node_modules/@wavs/solidity` remapping
https://github.com/Lay3rLabs/wavs-foundry-template/blob/main/remappings.txt#L2C1-L2C35

[...]

PR #67, opened 2025-04-08
feat: npm compatible 

reecepbcups:
ref: https://github.com/Lay3rLabs/wavs-middleware/issues/32

https://github.com/Lay3rLabs/wavs-foundry-template/pull/119

While upgrading the template to this repo it required this small change, the repo must have a package.json. nothing else is required

https://github.com/Lay3rLabs/wavs-foundry-template/blob/92b50949ae56ea779388758a48f8cb5c42ec8902/package.json#L28

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "wavs-middleware",
+  "version": "1.0.0",
+  "description": "WAVS Middleware",
+  "main": "index.js",
+  "author": "Lay3rLabs Team",
+  "license": "MIT"
+}
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "wavs-middleware",
   "version": "1.0.0",
   "description": "WAVS Middleware",
-  "main": "index.js",
   "author": "Lay3rLabs Team",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "wavs-middleware",
+  "version": "1.0.0",
```
ethanfrey: Please use "0.4.0-alpha.1" as that is the current version.
We should track WAVS tags, probably "0.4.0-beta" soon

ethanfrey (commented):
Like the idea if it helps you. Nit on the version used.

package.json
reecepbcups: resolved: e394023a6fd46e1aeac74e8cc7c8ff197d9ec564

RonTuretzky (commented):
not sure how this will be used upstream

ethanfrey:
LGTM. Ron you can sync with Reece on this. 
But first read the attached issue where we had discussed the workflows.
https://github.com/Lay3rLabs/wavs-middleware/issues/32

PR #68, opened 2025-04-08
fix: docker build latest git submodules if none exists

reecepbcups:
## Summary

- Before the docker image would __only__ build if you had run `git submodule update --init --recursive` before `docker build`. To ensure new clones (and gh actions in the future), a git modules update and copy over is performed. 
- While `git-deps` uses ubuntu:latest, only the files are copied over. This results in the exact same image size as before, no extra overhead.
- git-deps is first so it can be cached & does not require submodule downloads on all subsequent builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,11 @@
 FROM ubuntu:latest AS git-deps
 RUN apt update && apt install -yq git
-COPY . /tmp/
-RUN cd /tmp && git submodule update --init --recursive
+RUN mkdir -p /tmp/contracts/lib
+COPY .gitmodules contracts/lib /tmp/
+WORKDIR /tmp
+# a git env required to submodule pull
+RUN git init
+RUN git submodule update --init --recursive
 
 FROM rust AS builder
 COPY operator /wavs/operator/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,9 @@
-FROM rust AS builder
+FROM ubuntu:latest AS git-deps
+RUN apt update && apt install -yq git
+COPY . /tmp/
```
ethanfrey: Can you simplify this to enable better caching?

Write now this build phase is run each time any file changes.
You see how the rust builder only re-runs when files in operator change?
`COPY .` is a big pull. Maybe there is much less info needed to get the submodules, so it only re-runs when the remote references are run. Can you check?

ethanfrey (commented):
Good stuff in general, but can you optimize this build phase to cache better?

Dockerfile
reecepbcups: yes, resolved with 83d18f93430801637c5dc9fd448ef8e5ab9fb41f

Only submodule updates if the .gitmodules or contracts/lib directory is changed.

RonTuretzky (approved):
lgtm

ethanfrey (approved):
Much nicer. Thanks!

PR #69, opened 2025-04-09
testnet compatiblity 

hudsonhrh:
Working with the deploy.sh and set_service_uri.sh

Failing at register.sh

Add FUNDED_KEY to env and update the strategies 
LST_STRATEGY_ADDRESS=0x7D704507b76571a51d9caE8AdDAbBFd0ba0e63d3


ethanfrey:
Thank you for this. Could you link the issue(s)?

I prefer one PR for issue, so we can merge faster. Like just the layer to wavs renaming as one.

Also, I think this would be much easier to test if the local fork was running the same code as testnet, which I brought up elsewhere with @RonTuretzky 

Finally, part of the full testnet I would expect some docs mainly. How do we follow up on this

docker/deploy.sh
```diff
@@ -244,7 +244,6 @@ if [ "$DEPLOY_ENV" = "TESTNET" ]; then
         exit 1
     fi
 else
-    LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL:-http://localhost:8545}
```
ethanfrey: please don't remove this.
it provides a sane default

docker/env.example
```diff
@@ -4,7 +4,7 @@ RPC_URL=https://holesky.drpc.org
 DEPLOY_ENV=LOCAL
 CHAIN_ID=17000
 ETHERSCAN_API_KEY=foobar
-LOCAL_ETHEREUM_RPC_URL=http://localhost:8545
```
ethanfrey: why changing this? an issue using `--network host`?

If you are on mac, please read this to set it up properly https://docs.wavs.xyz/tutorial/2-setup#docker

ethanfrey:
I pulled most of your work (which was on the rename-wavs) branch and added a minor cleanup to it and pushed to #70 

This should be a much smaller PR focused on testnet deploy.

As I mentioned to @RonTuretzky a great first step would be to make the local fork and testnet deploy take the same path. This is mentioned in https://github.com/Lay3rLabs/wavs-middleware/blob/dev/docker/register.sh#L59 and a number of other TODOs in register.sh (and the other shell scripts).

Let's get it as close a possible on the testnet fork, and then I think the actual testnet deploy is mostly documentation

RonTuretzky:
Working e2e on testnet
 [run for ref](https://holesky.etherscan.io/address/0xa29004e593f6e2490ee18aa2262729f487ff4a49)

operator/rust/crates/operator/src/register_wavs_operator.rs
```diff
@@ -18,18 +18,16 @@ use once_cell::sync::Lazy;
...
-    match option_env!("TESTNET_RPC_URL") {
-        Some(url) => url,
-        None => "http://localhost:8545",
-    }
+fn get_rpc_url() -> String {
+    env::var("TESTNET_RPC_URL").unwrap_or_else(|_| "http://localhost:8545".to_string())
 }
-pub const ANVIL_RPC_URL: &str = get_rpc_url();
+
+static ANVIL_RPC_URL: Lazy<String> = Lazy::new(|| get_rpc_url());
```
ethanfrey, line 25: Ah, nice... the last version was pulling the env option compile time not runtime.

docker/set_service_uri.sh
```diff
@@ -22,6 +22,17 @@ else
     LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL:-http://localhost:8545}
 fi
 
+# Read the deployer private key from file
```
ethanfrey, line 25: Good we reuse this here.

I still think we might want to pass all the key in for any real deploy (rather than auto-generate), but at least we are the same in both scripts now

ethanfrey (commented):
Looks better, but not sure if it closes an issue

ethanfrey (approved):
Good to merge.

PR #70, opened 2025-04-09
Rename wavs

ethanfrey:
Pulling in all the renaming parts of #69 and doing minor cleanup to merge


operator/rust/crates/operator/src/register_wavs_operator.rs
```diff
@@ -21,7 +21,7 @@ use std::{env, str::FromStr};
 pub const fn get_rpc_url() -> &'static str {
     match option_env!("TESTNET_RPC_URL") {
         Some(url) => url,
-        None => "http://localhost:8545",
```
ethanfrey: You see why I want this a bash script, so we can configure like everywhere else

RonTuretzky: agreed

PR #71, opened 2025-04-09
Undocumented env config

RonTuretzky:
binding default variable to be docker internal host for docker compatibility, may be problematic later but experiencing an error on dev without this patch


RonTuretzky (commented):
closing as this will break linux compat

PR #72, opened 2025-04-09
61 clean up erc20 usage in layermiddlewaredeployer

### LINKED ISSUE #61: Clean up ERC20 usage in LayerMiddlewareDeployer

RonTuretzky:
closes #61


contracts/script/WavsMiddlewareDeployer.s.sol
```diff
@@ -35,9 +35,8 @@ contract WavsMiddlewareDeployer is Script, IECDSAStakeRegistryTypes {
         vm.label(deployer, "Deployer");
 
         coreDeployment = ReadCoreLib.readDeploymentJson("deployments/core/", block.chainid);
-       
-        token = new ERC20Mock();
-        helloWorldStrategy = IStrategy(StrategyFactory(coreDeployment.strategyFactory).deployNewStrategy(token));
+        // This is the strategy address for seth on holesky 
+        helloWorldStrategy = IStrategy(0x7D704507b76571a51d9caE8AdDAbBFd0ba0e63d3);
```
ethanfrey: It would be good to make this configurable, take it in as an env variable.
It only works on holesky and this script should be generic over deploy locations.

Just add this to the env.example file and pass it in

ethanfrey (commented):
Minor cleanup request

PR #73, opened 2025-04-10
Add query for operators

ethanfrey:
Add queries to see if any operators are actually registered on the contract.

I ran the local fork deploy and this show no weight, 0 threshold, and no operators registered.

Something is wrong here. This is why I was pushing for the local and testnet to be the same


ethanfrey:
@RonTuretzky can you please verify I am doing this correctly?

Cuz nothing seems to be set on the actual staking registry from the deployment file

ethanfrey:
I merged all open PRs and rebased this on top. I now get some events, and properly see the AVS KEYS that were registered. But all values (weight, threshold, etc) are 0. Here are the results:

```
=== ECDSA Stake Registry Status ===
Contract Address: 0x0657fe1a890da6e517c9bf91f23b310a8bfd6f2e

=== Quorum Information ===
Total Weight: 0
Threshold Weight: 0

=== Registered Operators ===
Querying events from block 3646717 to 3648717
Found operator: 0x3464592915269a1dbdd65b8e3452011e43d50c59
Found operator: 0x8181d98801bb19150e2d5a42900815402ba29953

=== Operator Weights ===
Operator 0x3464592915269a1dbdd65b8e3452011e43d50c59 weight: 0
Operator 0x8181d98801bb19150e2d5a42900815402ba29953 weight: 0
```

Those two addresses match what I got from `cast wallet addr $AVS_KEY` for the two runs I made registering operators.
This is now using all of the "testnet fixes"

docker/list_operator.sh
```diff
@@ -0,0 +1,78 @@
...
+# Get total weight and threshold
+echo -e "\n=== Quorum Information ==="
+TOTAL_WEIGHT=$(cast call "$ECDSAStakeRegistryAddress" "getLastCheckpointTotalWeight()(uint256)" --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
+THRESHOLD_WEIGHT=$(cast call "$ECDSAStakeRegistryAddress" "getLastCheckpointThresholdWeight()(uint256)" --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
+echo "Total Weight: $TOTAL_WEIGHT"
+echo "Threshold Weight: $THRESHOLD_WEIGHT"
+
+# Get current block height and calculate range
+LATEST_BLOCK=$(cast block-number --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
+FROM_BLOCK=$((LATEST_BLOCK - 2000))
```
dakom, line 40: I'm assuming this is always valid for our targets... but if hitting some young chain for some reason, could be problematic? not a blocker imho

docker/list_operator.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+
+# -x echos all lines for debug
+# set -x
+
+set -o errexit -o nounset -o pipefail
+command -v shellcheck >/dev/null && shellcheck "$0"
```
dakom, line 7: I'm not so savvy on shell scripting, relied on ChatGPT to tell me that this was good and smart 👍

docker/list_operator.sh
```diff
@@ -0,0 +1,78 @@
...
+ECDSAStakeRegistryAddress=$(cat /root/.nodes/avs_deploy.json | jq -r '.addresses.stakeRegistry')
+if [ -z "$ECDSAStakeRegistryAddress" ]; then
+    echo "Error: failed to read ECDSAStakeRegistry from /root/.nodes/avs_deploy.json"
+    exit 1
+fi
+
+echo "=== ECDSA Stake Registry Status ==="
+echo "Contract Address: $ECDSAStakeRegistryAddress"
+
+# Get total weight and threshold
```
dakom, line 31: hmm... yeah, it would be nice to optionally have this hit the service manager contract for the sake of testing that everything is wired up correctly, but let's definitely get this working first and only then consider expanding features

docker/list_operator.sh
```diff
@@ -0,0 +1,78 @@
...
+
+if [ -z "$OPERATOR_EVENTS" ]; then
+    echo "No OperatorRegistered events found in the specified block range."
+    exit 1
+fi
+
+
+# It is the second line after topics,
+# and looks like 0x0000000000000000000000003464592915269a1dbdd65b8e3452011e43d50c59
+RAW_OPS=$(echo "$OPERATOR_EVENTS" | grep -A2 'topics:' | grep '0x00000000000000')
```
dakom, line 55: erm, I don't know how ethereum does this, but if it works it works! I do wonder if `--json` would make it easier to understand, but no biggie

dakom (approved):
Waiting till I can follow the README.md to properly run, but I read through this and checked some parts I'm less familiar with via ChatGPT. Left some comments but all looks good!

dakom:
Works for me. I get `0` values, as noted. Opened https://github.com/Lay3rLabs/wavs-middleware/issues/78 to track that

PR #79, opened 2025-04-10
Update README.md

dakom:
Just to emphasize a step I missed :)

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -36,6 +36,10 @@ anvil --fork-url $RPC_URL --host 0.0.0.0 --port 8545
 
 **Run all the following scripts in the `docker/` directory.**
 
+```bash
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -34,7 +34,7 @@ anvil --fork-url $RPC_URL --host 0.0.0.0 --port 8545
 
 ## Deploy
 
-Run all the following scripts in the `docker/` directory.
+**Run all the following scripts in the `docker/` directory.**
```
reecepbcups, line 37: i have missed this twice too.

Should we just put a `cd docker/` as one of the commands?

dakom: makes sense to me :)

reecepbcups (approved):

PR #95, opened 2025-04-30
Yolo

### LINKED ISSUE #55: POA Opset (ECDSAStakeRegistry compatible)
Motivation: 
Being able to run wavs with an operator set that does not have to go through any eigenlayer registration for usecases that don't want to configure this/ are bootstrapping / need go to market. 

- [x] implements all interfaces  expected from WAVS 
- [x] Initialize with an operator set (address array) 
- [x] implement signature verification logic (we can keep blocknum inside but accept any blocknum)

### LINKED ISSUE #80: Remove breadchain Eigenlayer Middleware fork
Create new middleware fork
Update submodules

JakeHartnell:
- Closes #80
- Closes #55 (might make sense to put it in a separate repo?)

Reorgized the codebase to be a bit more of a standard forge repo. Was able to successfully register and activate an operator on Holesky testnet. Key thing here was remembering to update stakes on Eigenlayer after registering an operator and staking LSTs.


ethanfrey (commented):
I guess this is a WIP?

I'd love to cover the redesign purpose. I see generally you are moving from bash to solidity. I do agree that moving off bash would be good mid-term, but trying to keep things stable. I'm not sure solidity scripts are the best - we could also write deployment scripts in Rust or Javascript. David wrote some in Rust, Milan in Javascript before. Both seem more mature languages than Solidity.

Happy to discuss the architecture decisions more. And if there is more stuff in here I should be aware of (beyond porting bash to forge solidity deploy scripts)

PR #97, opened 2025-05-07
Reorg solidity files to separate mocks, docker, etc

### LINKED ISSUE #40: Massive repo trimming
We want this to be the solidity central for all wavs stuff.
We don't want to have rust or js stuff here (just the foundry cli tools in docker and the scripts)
Time for a massive reorg here.

- [x] Clean out everything we don't need for the basic deploy scripts
- [ ] Reorganize top-level directories

Desired directory structure:

* docker
  * scripts for our docker image
* interfaces
  * Expose solidity interfaces for common pieces, to be used also by mocks and wavs calling it
* eigenlayer_1_1
  * all the actual contracts we are using now
* mock
  * some dummy contracts we use for WAVS CI testing

Later we will add things like `eigenlayer_1_4` to store versions of those contracts

ethanfrey:
Closes #40 

After discussion with @dakom reorganizes the solidity files to match our [discussion here](https://github.com/Lay3rLabs/wavs-middleware/issues/40#issuecomment-2834576132), except `lib` is top level rather than inside `eigenlayer/lib` (I kept getting compile errors).

Mocks are separated out and hope this makes the WAVS (and wavs-foundry-template) build process easier

I have manually tested both building the solidity code as running the updated docker scripts.

I would like approval from @dakom that this works with WAVS
I would like approval from @Reecepbcups that this doesn't break the wavs-foundry-template


reecepbcups (approved):
Confirming I got a fresh docker build based on 82c96d8ed1e972feea32be7019e69c8fdadec396 working in the template without any changes. _both the main README & multi example w/ aggregator_

dakom (approved):
everything works when downloading/rebuilding from WAVS :)

also confirmed that `forge build --root ./contracts ./mocks` works

PR #98, opened 2025-05-08
holesky testnet 

reecepbcups:
I am build thing this on machine for now. pairs with https://github.com/Lay3rLabs/wavs-foundry-template/pull/147

( https://github.com/Lay3rLabs/wavs-foundry-template/pull/166 is the main refactor using this as of May 20th)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes --entryp
 
 ## Deploy Testnet
 
-Same as the local deploy, but add `TESTNET_RPC_URL` to the .env and change `DEPLOY_ENV` to `"TESTNET"` and make sure the `PRIVATE_KEY` is actually funded on testnet
+Same as the local deploy, but add `TESTNET_RPC_URL` to the .env and change `DEPLOY_ENV` to `"TESTNET"` and make sure the `FUNDED_KEY` is actually funded on testnet
 
 ## References
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes --entryp
 
 ## Deploy Testnet
 
-Same as the local deploy, but add `TESTNET_RPC_URL` to the .env and change `DEPLOY_ENV` to `"TESTNET"` and make sure the `FUNDED_KEY` is actually funded on testnet
+Same as the local deploy, but add `TESTNET_RPC_URL` to the .env and change `DEPLOY_ENV` to `"TESTNET"` and make sure the `PRIVATE_KEY` is actually funded on testnet
 
 ## References
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ AVS_KEY=$(cast wallet new --json | jq -r '.[0].private_key')
 # This will show the address, so you can confirm it was properly added when listing operators
 cast wallet addr --private-key "$AVS_KEY"
 
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY"
+docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY" "0.01ether"
 ```
 
 List Operators:
--- a/README.md
+++ b/README.md
@@ -67,14 +67,22 @@ AVS_KEY=$(cast wallet new --json | jq -r '.[0].private_key')
 # This will show the address, so you can confirm it was properly added when listing operators
 cast wallet addr --private-key "$AVS_KEY"
 
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY" "0.01ether"
+export WAVS_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' .nodes/avs_deploy.json)
+export STAKE_REGISTRY_ADDRESS=$(jq -r '.addresses.stakeRegistry' .nodes/avs_deploy.json)
+
+docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes \
+   -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
+   -e STAKE_REGISTRY_ADDRESS=${STAKE_REGISTRY_ADDRESS} \
+   --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY" "0.01ether"
 ```
 
 List Operators:
 
 ```bash
 # View stake registry status, including registered operators and their weights
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes --entrypoint /wavs/list_operator.sh wavs-middleware
+docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes \
+   -e STAKE_REGISTRY_ADDRESS=${STAKE_REGISTRY_ADDRESS} \
+   --entrypoint /wavs/list_operator.sh wavs-middleware
 ```
 
 ## Deploy Testnet
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -37,9 +37,9 @@ if [ -z "$METADATA_URI" ]; then
     exit 1
 fi
 
-# Check if FUNDED_KEY is provided
-if [ -z "$FUNDED_KEY" ]; then
-    echo "Error: FUNDED_KEY environment variable must be set"
+# Check if PRIVATE_KEY is provided
+if [ -z "$PRIVATE_KEY" ]; then
+    echo "Error: PRIVATE_KEY environment variable must be set"
     exit 1
 fi
 
@@ -74,7 +74,7 @@ create_operator_set() {
           "cast s '$WavsServiceManagerAddress' \
              'createOperatorSets((uint32,address[])[])' \
              '[($set_id,[$LST_STRATEGY_ADDRESS])]' \
-             --private-key '$FUNDED_KEY' \
+             --private-key '$PRIVATE_KEY' \
              --rpc-url '$LOCAL_ETHEREUM_RPC_URL' > /dev/null 2>&1"
     else
         execute_transaction "created operator set $set_id with 1 strategy" \
@@ -100,7 +100,7 @@ update_avs_registrar() {
           "cast s '$WavsServiceManagerAddress' \
              'setAVSRegistrar(address)' \
              '$avsRegistrarAddress' \
-             --private-key '$FUNDED_KEY' \
+             --private-key '$PRIVATE_KEY' \
              --rpc-url '$LOCAL_ETHEREUM_RPC_URL' > /dev/null 2>&1"
     else
         execute_transaction "set up the AVSRegistrar through WavsServiceManager" \
@@ -144,7 +144,7 @@ update_metadata_url() {
       execute_transaction "updated AVS metadata URI" \
         "cast s '$serviceManagerAddress' 'updateAVSMetadataURI(string)' \
          '$METADATA_URI' \
-         --private-key '$FUNDED_KEY' \
+         --private-key '$PRIVATE_KEY' \
          --rpc-url '$LOCAL_ETHEREUM_RPC_URL' > /dev/null 2>&1"
   else
       execute_transaction "updated AVS metadata URI" \
@@ -169,7 +169,7 @@ update_quorum_config() {
              'updateQuorumConfig(((address,uint96)[]),address[])' \
              '([$combined_strategies])' \
              '[]' \
-             --private-key '$FUNDED_KEY' \
+             --private-key '$PRIVATE_KEY' \
              --rpc-url '$LOCAL_ETHEREUM_RPC_URL' > /dev/null 2>&1"
     else
         execute_transaction "update quorum config" \
@@ -202,7 +202,7 @@ update_minimum_weight() {
              'updateMinimumWeight(uint256,address[])' \
              '$minimumWeight' \
              '[]' \
-             --private-key '$FUNDED_KEY' \
+             --private-key '$PRIVATE_KEY' \
              --rpc-url '$LOCAL_ETHEREUM_RPC_URL' > /dev/null 2>&1"
     else
         execute_transaction "update minimum weight" \
@@ -228,7 +228,7 @@ deploy_consumer_contract() {
     offchainMessageConsumerAddress=$(
         forge create --json \
           -r "$LOCAL_ETHEREUM_RPC_URL" \
-          --private-key "$FUNDED_KEY" \
+          --private-key "$PRIVATE_KEY" \
           --broadcast src/example-contracts/OffchainMessageConsumer.sol:OffchainMessageConsumer \
           --constructor-args "$stakeRegistryAddress" \
         | jq -r '.deployedTo'
@@ -253,10 +253,9 @@ deploy_consumer_contract() {
 #############################################
 
 mkdir -p ~/.nodes
-deployer_public_key=$(cast wallet address "$FUNDED_KEY")
-echo "PRIVATE_KEY=$FUNDED_KEY" >> contracts/.env
-echo "$FUNDED_KEY" > ~/.nodes/deployer
-source contracts/.env
+deployer_public_key=$(cast wallet address "$PRIVATE_KEY")
+echo "PRIVATE_KEY=$PRIVATE_KEY" >> contracts/.env
+echo "$PRIVATE_KEY" > ~/.nodes/deployer
 
 if [ "$DEPLOY_ENV" = "TESTNET" ]; then
     LOCAL_ETHEREUM_RPC_URL="$TESTNET_RPC_URL"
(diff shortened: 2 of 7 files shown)


docker/register.sh
```diff
@@ -103,14 +103,15 @@ setup_operator() {
...
-    
+
     if [ "$DEPLOY_ENV" = "TESTNET" ]; then
         # TODO: remove this and replace with a check the AVS key has a balance
-        cast s "$public_key" --value 50000000000000000 --private-key "$FUNDED_KEY" -r "$LOCAL_ETHEREUM_RPC_URL" > /dev/null 2>&1
-        if [ $? -ne 0 ]; then
-            echo "Error: Failed to give operator $index balance"
-            exit 1
-        fi
+        # cast s "$public_key" --value 50000000000000000 --private-key "$FUNDED_KEY" -r "$LOCAL_ETHEREUM_RPC_URL" > /dev/null 2>&1
```
reecepbcups: this is needed for testnet, if balance ignore send

docker/register.sh
```diff
@@ -121,27 +122,29 @@ setup_operator() {
...
-    cast send "$LST_CONTRACT_ADDRESS" "$MINT_FUNCTION" "$public_key" "0x0000000000000000000000000000000000000000" \
-        --private-key "$private_key" \
-        --value 10000000000000000 \
-        --rpc-url "$LOCAL_ETHEREUM_RPC_URL" > /dev/null 2>&1
-    if [ $? -ne 0 ]; then
-        echo "Error: Failed to mint LST for $ADDRESS"
-        exit 1
-    fi
+    # TODO: Need to be able to change the amount we stETH stake relative to their balance. TRight now it hands for me?
+    AMOUNT=10000
```
reecepbcups: expose via ENV variable

docker/register.sh
```diff
@@ -121,27 +122,29 @@ setup_operator() {
...
+    # MINT_FUNCTION="submit(address _referral)"
+    # cast send "$LST_CONTRACT_ADDRESS" "$MINT_FUNCTION" "$public_key" "0x0000000000000000000000000000000000000000" \
+    #     --private-key "$private_key" \
+    #     --value ${AMOUNT} \
+    #     --rpc-url "$LOCAL_ETHEREUM_RPC_URL" > /dev/null 2>&1
+    # if [ $? -ne 0 ]; then
+    #     echo "Error: Failed to mint LST for $ADDRESS"
+    #     exit 1
+    # fi
```
reecepbcups: this is pending issue [reported in slack atm](https://lay3r.slack.com/archives/C07SZ6G0T35/p1746722108467859?thread_ts=1746721960.445869&cid=C07SZ6G0T35)

docker/list_operator.sh
```diff
@@ -37,7 +37,7 @@ echo "Threshold Weight: $THRESHOLD_WEIGHT"
 
 # Get current block height and calculate range
 LATEST_BLOCK=$(cast block-number --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
-FROM_BLOCK=$((LATEST_BLOCK - 2000))
```
ethanfrey, line 40: why this change? ideally we want all registrations ever, 2000 was small, 900 is smaller.

docker/register.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # -x echos all lines for debug
-# set -x
+set -x
```
ethanfrey: Debug remnant, please remove before merge

docker/register.sh
ethanfrey: This is cuz public_key and FUNDED_KEY are the same right now?

The comment says to replace with a check, but it is just removed. I guess something more is needed?

docker/register.sh
```diff
@@ -121,27 +122,29 @@ setup_operator() {
 
     echo "Using LST_CONTRACT_ADDRESS: $LST_CONTRACT_ADDRESS"
     echo "Using LST_STRATEGY_ADDRESS: $LST_STRATEGY_ADDRESS"
-    
+
     # TODO: is this write? need proper LST addr setup in the .env file
+    # TODO: Need to be able to change the amount we stETH stake relative to their balance. TRight now it hands for me?
+    AMOUNT=10000 # TODO: make this a value set via cli args / env variables to override
```
ethanfrey: Good idea. But let's make the default higher. Like 0.01ETH at least. (Which would  be a 1 with 16 zeros)

ethanfrey (commented):
Some minor comments, let me know when ready for final review

docker/list_operator.sh
reecepbcups: ETH RPC nodes limited to 1000 (999 including latest block) with multiple RPC provides i tried on holesky

We might require even smaller for mainnet, not sure.

docker/register.sh
reecepbcups: yea we need a check: if bal is 0 faucet funds with the funded key. I just have not written the logic yet for the proper solution. I just faucet manually in the template for now

(right now it would fail because I was testing with way less than 0.5ether due to the testnet drain pain)

docker/register.sh
```diff
@@ -121,27 +122,29 @@ setup_operator() {
 
     echo "Using LST_CONTRACT_ADDRESS: $LST_CONTRACT_ADDRESS"
     echo "Using LST_STRATEGY_ADDRESS: $LST_STRATEGY_ADDRESS"
-    
+
     # TODO: is this write? need proper LST addr setup in the .env file
+    # TODO: Need to be able to change the amount we stETH stake relative to their balance. TRight now it hands for me?
```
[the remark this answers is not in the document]
reecepbcups: will do this when I push latest

docker/register.sh
reecepbcups: set as configurable now when calling `/wavs/register.sh`

docker/register.sh
```diff
@@ -25,59 +25,66 @@ if [ -z "$LST_STRATEGY_ADDRESS" ]; then
...
+    # if [ -z "$stake_registry_address" ]; then
+    #     echo "Error: Failed to read StakeRegistry from /root/.nodes/avs_deploy.json"
+    #     exit 1
+    # fi
+    local service_manager_address=0x542ea0be029d35a1bc83f3046dd753dbc74a26fb
+    # local service_manager_address=$(cat /root/.nodes/avs_deploy.json | jq -r '.addresses.WavsServiceManager')
+    # if [ -z "$service_manager_address" ]; then
+    #     echo "Error: Failed to read WavsServiceManager from /root/.nodes/avs_deploy.json"
+    #     exit 1
+    # fi
```
reecepbcups: these should be passed as arguments imo. not really persistent .env related, just temp 1 offs

docker/register.sh
```diff
@@ -174,7 +198,7 @@ setup_operator() {
         echo "Error: Failed to register operator $public_key to operator sets"
         exit 1
     fi
```
reecepbcups, line 203: This entire section of submit -> registerForOperatorSets needs to be ignored if the user is already in the operator set and / or they already have LST

Also `register_operator_with_avs` was throwing to dev/null and was impossible to debug without removing. we should really show as much as possible to the user
)side note: register_operator_with_avs failed bc it was trying to read from a file the operator did not have access too, only the deployer)

docker/register.sh
ethanfrey: Makes sense. Please update the script and the readme on how to use it

ethanfrey: I mean… they should not be hardcoded,  it we can’t expect the deploy json on the operator machines.

Agree they should be env vars

ethanfrey:
Let me know if you are ready for a review. Or if there are any blockers I can help with

reecepbcups:
not ready quite yet

Current status is in: https://github.com/Lay3rLabs/wavs-foundry-template/pull/162 & https://github.com/Lay3rLabs/wavs-foundry-template/pull/166. Just finished up the last remaining local devEx improvement with the local warg (OCI v1) server.

This work here has been blocked on me with that. I had to know the [ideal flow](https://docs.google.com/document/d/1RKWQHckHPGzGu6cRc9pQQEyXEWX8WnkXoHQ5yhBLmYU/edit?tab=t.0) before I could make sure the middleware match those requirements. so far looks like it did match 1 for 1, biggest risk now resolved in the template

I am aiming to have a working state realistically tomorrow morning / afternoon for the full flow, matching local & testnet deploys :D 

Will ping for review & Jakub to run through once I have that done

README.md
```diff
@@ -67,7 +67,7 @@ AVS_KEY=$(cast wallet new --json | jq -r '.[0].private_key')
 # This will show the address, so you can confirm it was properly added when listing operators
 cast wallet addr --private-key "$AVS_KEY"
 
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY"
+docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY" "0.01ether"
```
Reecepbcups, line 70: note

```
docker run --rm --network host \
		-e WAVSServiceManagerAddress=${WAVSServiceManagerAddress} \
		-e StakeRegistryAddress=${StakeRegistryAddress} \
		--env-file ${ENV_FILE} \
		-v ./.nodes:/root/.nodes \
		--entrypoint /wavs/register.sh ${MIDDLEWARE_DOCKER_IMAGE} "${AVS_PRIVATE_KEY}" "0.001ether"
```

either need to update the env file or reference this like I do in the template

docker/register.sh
```diff
@@ -25,9 +25,13 @@ if [ -z "$LST_STRATEGY_ADDRESS" ]; then
...
     exit 1
 fi
-WAVSServiceManagerAddress=$(cat /root/.nodes/avs_deploy.json | jq -r '.addresses.WavsServiceManager')
+
 if [ -z "$WAVSServiceManagerAddress" ]; then
-    echo "Error: failed to read WavsServiceManager from /root/.nodes/avs_deploy.json"
+    echo "Error: WAVSServiceManagerAddress is not set in the environment variables (tip: grab from .nodes/avs_deploy.json)."
+    exit 1
+fi
+if [ -z "$StakeRegistryAddress" ]; then
```
reecepbcups: migrate to all caps env var or leave as is? (this is user facing, but not a hardcoded value)

docker/register.sh
reecepbcups: someone was snipping my eth in the mempool of an anvil account lmao

docker/register.sh
reecepbcups: logic written in a later commit (polls for balance -gt 0 request check

docker/register.sh
reecepbcups: moved to env values

docker/register.sh
reecepbcups: bash convention for external variables like this is still UPPER_CASE

reecepbcups:
Ready for review now. -- I have run through this **many** times (local & testnet) and Jakub ran through it on holesky successfully first try in the template

ethanfrey:
Great, I will run through this tonight and approve/merge if I can run through it properly.
Any comments on docs will go into new issues, I really want to get this in if it works.

README.md
```diff
@@ -67,14 +67,22 @@ AVS_KEY=$(cast wallet new --json | jq -r '.[0].private_key')
 # This will show the address, so you can confirm it was properly added when listing operators
 cast wallet addr --private-key "$AVS_KEY"
 
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes  --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY"
+export WAVS_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' .nodes/avs_deploy.json)
```
ethanfrey, line 70: Great stuff. Pull out the variables we actually need from the json so we can provide this to the operators.

Ideally, we would have a separate OPERATORS.md or such in a project that contained this, and includes the actual service manager address. But this is great for testing

README.md
```diff
@@ -67,14 +67,22 @@ AVS_KEY=$(cast wallet new --json | jq -r '.[0].private_key')
...
+export STAKE_REGISTRY_ADDRESS=$(jq -r '.addresses.stakeRegistry' .nodes/avs_deploy.json)
```
ethanfrey, line 71: We can query this from the service manager.
Ideally we make just one variable to configrue everything (service manager, since we get service.json url there as well), and query the rest from the blockchain.

README.md
```diff
@@ -67,14 +67,22 @@ AVS_KEY=$(cast wallet new --json | jq -r '.[0].private_key')
...
+   --entrypoint /wavs/register.sh wavs-middleware "$AVS_KEY" "0.01ether"
 ```
 
 List Operators:
 
 ```bash
 # View stake registry status, including registered operators and their weights
-docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes --entrypoint /wavs/list_operator.sh wavs-middleware
+   -e STAKE_REGISTRY_ADDRESS=${STAKE_REGISTRY_ADDRESS} \
```
ethanfrey, line 84: As above, I would just pass in service manager address and query the rest from there

docker/deploy.sh
```diff
@@ -296,6 +295,21 @@ fi
 
 echo "Deployer address: $deployer_public_key configured for $DEPLOY_ENV environment"
 
+FUNDED_KEY_BAL=$(cast balance ${deployer_public_key} --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
```
ethanfrey, line 298: nice to ensure we actually have funds on testnet

docker/register.sh
```diff
@@ -36,60 +40,57 @@ register_operator_with_avs() {
...
-    cast send "$stake_registry_address" \
-        "registerOperatorWithSignature((bytes,bytes32,uint256),address)" \
-        "($signature,$salt,$expiry)" "$public_key" \
-        --private-key "$private_key" \
-        --rpc-url "$LOCAL_ETHEREUM_RPC_URL" 
-    if [ $? -eq 0 ]; then
-        echo "Successfully registered operator $public_key with AVS"
+
+    local operatorRegistered=$(cast call "$STAKE_REGISTRY_ADDRESS" "operatorRegistered(address)(bool)" "$public_key" --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
+    if [ "$operatorRegistered" = "false" ]; then
```
ethanfrey, line 64: nice check, good error handling if they are already registered

docker/register.sh
```diff
@@ -36,60 +40,57 @@ register_operator_with_avs() {
...
+        echo "Registering operator with signature..."
+        cast c "$STAKE_REGISTRY_ADDRESS" \
+            "registerOperatorWithSignature((bytes,bytes32,uint256),address)" \
```
ethanfrey, line 67: Why is this twice? Once with `cast c` and once with `cast send`? I assume that is to dry run it, but why do you need to do this now you have checked for operator registered?

docker/register.sh
```diff
@@ -36,60 +40,57 @@ register_operator_with_avs() {
...
+
+        cast send "$STAKE_REGISTRY_ADDRESS" \
+            "registerOperatorWithSignature((bytes,bytes32,uint256),address)" \
+            "($signature,$salt,$expiry)" "$public_key" \
+            --private-key "$private_key" \
+            --rpc-url "$LOCAL_ETHEREUM_RPC_URL"
+        if [ $? -eq 0 ]; then
+            echo "Successfully registered operator $public_key with AVS"
+        else
+            echo "Error: Failed to register operator with AVS"
```
ethanfrey, line 80: This will never be hit, as we have `set -o errexit` on top, which means the script will terminate as soon as any line returns a non-zero error code.

You could do something like `cast send ... || (echo "Error: Failed to register" && exit 1)` to have the same effect

docker/register.sh
```diff
@@ -103,14 +104,18 @@ setup_operator() {
         exit 1
     fi
 
-    
     if [ "$DEPLOY_ENV" = "TESTNET" ]; then
-        # TODO: remove this and replace with a check the AVS key has a balance
-        cast s "$public_key" --value 50000000000000000 --private-key "$FUNDED_KEY" -r "$LOCAL_ETHEREUM_RPC_URL" > /dev/null 2>&1
+        balance=$(cast balance "$public_key" --rpc-url "$LOCAL_ETHEREUM_RPC_URL")
```
ethanfrey, line 108: Could we really rename public key to address? Cuz that is what it is (see line 92)

docker/register.sh
```diff
@@ -121,70 +126,96 @@ setup_operator() {
...
 
-    cast send "$DELEGATION_MANAGER_ADDRESS" \
-        "registerAsOperator(address,uint32,string)" \
-        "$public_key" 0 "foo.bar" \
-        --private-key "$private_key" \
-        --rpc-url "$LOCAL_ETHEREUM_RPC_URL" > /dev/null 2>&1
-    if [ $? -ne 0 ]; then
-        echo "Error: Failed to register as operator for $DELEGATION_MANAGER_ADDRESS"
-        exit 1
+    # If the operator has deposits, we don't need to do anything
```
ethanfrey, line 132: Great stuff, this figures out where in the registration sequence the operator is (do they have deposits, are they registered, do they have LST, do they have ETH)? And then just does the needed steps.

It may be a bit easier to do all the queries first (check where you are), then run the various send commands that are needed.

ethanfrey (approved):
Great stuff and please merge as is. I don't want to keep this out of dev branch any longer.

I would ask you to then make a follow up PR that addresses all these comments. Push that as a separate branch/PR, as this has been reviewed, manually tested, and approved. And then we can get a much smaller cleanup PR to give the fine tuning.

docker/register.sh
reecepbcups: Ahhh, this may be why the makefile never returns the actual error message in the template, just the return code.

yea will refactor to this or design, thanks for the tip here super helpful

docker/register.sh
reecepbcups: this was there before, removed and it just worked :+1:

PR #102, opened 2025-05-12
Trim down Service Manager interface

### LINKED ISSUE #100: Clean up ServiceManager interface
Related to https://github.com/Lay3rLabs/WAVS/issues/621

Remove these [two queries](https://github.com/Lay3rLabs/wavs-middleware/blob/dev/contracts/interfaces/IWavsServiceManager.sol#L24-L34)

Add another `InsufficientWeight` error that can also be thrown by the validate funciton

ethanfrey:
Closes #100 

Removes some methods that are not needed (should not be used) after https://github.com/Lay3rLabs/WAVS/issues/621


dakom (approved):
WAVS updated to work with this interface :)

Issue #39, opened 2025-03-26
Set Service Key


RonTuretzky:
### Motivation
Prevent operators from using a key that potentially controls billions of dollars for daily avs signing operations 

### Background 

This is implemented "naturally' in the [BLS](https://github.com/Layr-Labs/eigenlayer-middleware/tree/dev/src)  , we would like to add in the registration process a "declaration" of a "Signing key" by having the operator key sign the signing key , and verify that upon registration 

We need to maintain a mapping between the signing key and the operator key as we will need to verify threshold signatures using the signing keys instead of the original implementation (which uses the operator keys)

RonTuretzky:
Suggesting we postpone this as it's solved by the BLS migration naturally

ethanfrey:
We want this on the ECDSA contracts as well.

What is needed is a mapping from each operator key (that is known by eigenlayer and is delegated stake and received rewards) to an AVS-specific signing key. The AVS-key is known by the WAVS node and has no tokens, only used to validate the signatures on the packet that is verified by the ServiceManager.

The registration (and thus using the operator key) is only done once in deploy scripts.

This keeps the two keys separate, and also allows us to have one operator run multiple services on the same WAVS node with the same delegated stake, but different signing keys.

Ideally, the operator would be able to update the signing key later, but we need to maintain some history to allow for retroactive slashing.

ethanfrey:
Fixed between #120 and the bugfixes in #130 

Tested this now full-stack in https://github.com/Lay3rLabs/wavs-foundry-template/pull/178 and it works.

PR #120, opened 2025-05-26
39 set service key

endrit1991:
Fixed issue #39 

ECDSAStakeRegistry already has function registerOperatorWithSignature. It takes address of signing key that can be different than operator address (msg.sender).

I update register script to take extra arguments for AVS_SIGNING_ADDRESS and OPERATOR_KEY so they can be different.

Careful. ECDSAStakeRegistry.isValidSignature takes array of operator address. And array of signatures signed with signing key. This maybe confusing when calling from your code.

Please call 
git submodule update --recursive
when you check out to get improved dependency


ethanfrey:
Thank you for the work here and surprised that you found this functionality already worked. We had been informed this was only in BLS contract and not ECDSA.

The code you provided looks good and I have tested it manually, providing different signing addresses. This is a good step forward. However, I cannot test with WAVS currently, as it uses the same address for signing as for the operators array that you mention, as you mentioned

> Careful. ECDSAStakeRegistry.isValidSignature takes array of operator address. And array of signatures signed with signing key. This maybe confusing when calling from your code.

That said, the current upstream `ECDSAStakeRegistry` implementation doesn't meet our needs 100%. Managing two keys is unnecessary overhead for our WAVS node and adds confusion in the processing. I would ask you to make a PR in our https://github.com/Lay3rLabs/eigenlayer-middleware repo to update [`_checkSignatures`](https://github.com/Lay3rLabs/eigenlayer-middleware/blob/v1.4.0-testnet-holesky/src/unaudited/ECDSAStakeRegistry.sol#L422-L427) so that it takes an array of signers not an array of operators. Please ensure there are test cases for this, and existing test cases are upated.

The PR should target the `wavs-1.4` branch, which is currently at `v1.4.0-testnet-holesky` (the same as we currently use in wavs-middleware submodules)

Please link that PR to this one, and once I have merged the eigenlayer-middleware PR, you should update this one to reference that one in the submodules and I am happy to merge this.

endrit1991:
Okay, I work on the eigenlayer-middleware tomorrow as requested. Please give me write access.

endrit1991:
Please see PR https://github.com/Lay3rLabs/eigenlayer-middleware/pull/1

ethanfrey:
Tested this and looks good.
I will merge so the docker image is built, then test it out with template.
Once the template and WAVS works with this, then I will consider this fully done.

ethanfrey (approved):
Approved. And nice upstream changes on eigenlayer-middleware

Issue #90, opened 2025-04-23
Better handling of minimum weight


ethanfrey:
As of `v0.4.0-alpha.2` registering multiple operators has no affect on the weight, which is hardcoded to some super low value.

```bash
$ docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes --entrypoint /wavs/list_operator.sh wavs-middleware

=== ECDSA Stake Registry Status ===
Contract Address: 0x31b1939e53385631b0b27cda97840995319abf20

=== Quorum Information ===
Total Weight: 1980697118081818 [1.98e15]
Threshold Weight: 100

=== Registered Operators ===
Querying events from block 3718661 to 3720661
Found operator: 0xdfbb803341cb8d1e762b9a675e625f639a85b184
Found operator: 0x91d32fa5d0ba40d475a24714ca6f08e35e26b2c8

=== Operator Weights ===
Operator 0xdfbb803341cb8d1e762b9a675e625f639a85b184 weight: 990348559040909 [9.903e14]
Operator 0x91d32fa5d0ba40d475a24714ca6f08e35e26b2c8 weight: 990348559040909 [9.903e14]
```

We need a way of setting say 2/3 threshold and having the minimum weight update when new operators are registered.

This issue is not ready to start coding until we come up with a agreed design decision. Please add comments here explaining best practices on how to implement this, preferably links to documentation. 

I did look at [general eigenlayer developer docs](https://docs.eigenlayer.xyz/developers/Concepts/eigenlayer-contracts/middleware-contracts) and the [eigenlayer-middleware docs](https://github.com/Layr-Labs/eigenlayer-middleware/tree/dev/docs) and nothing mentions setting minimum stake. I know @RonTuretzky mentioned some sidecar process to run daily to sync the operator's stake from eigenlayer to the actual stake registry. It would be helpful to see docs on that as they might be relevant here as well.

Please add comments for any context on how best to approach this.

ethanfrey:
Discussion with Eigenlayer devs suggested that I could add this in the service manager. 

* The stake registry checkSignatures returns the total amount of stake that signed the packet.
* BWe also can query the total stake of all operators easily.
* The ServiceManager validate method can then apply any check (like 2/3 threshold) on the total amount
* This is apart from the minimum threshold that could mean the contract wouldn't be open at all until there was 10 ETH signing, for example. That is a fixed minimum, separate from the percentage quorum requested by this issue

[...]

PR #117, opened 2025-05-22
106 freeze operator set

endrit1991:
Fixed issue #106 

Update WavsAVSRegistrar so owner can freeze all new registrations and deregistrations. I add shell scripts for pause and unpause the registrar so it can be called from docker commands.

PR #124, opened 2025-05-27
90 better minimum weight

endrit1991:
Fixes issue #90 

This adds quorum check in Service Manager as requested. On creation quorum is set to 2/3, but it can be changed later by owner. This is only done after the ECDSAStakeRegistry check passes .

As requested I wrote unit tests for this change. I also added tests for the pause functionality as requested  in #117

### DIFF
--- a/contracts/eigenlayer/src/WavsServiceManager.sol
+++ b/contracts/eigenlayer/src/WavsServiceManager.sol
@@ -244,9 +244,8 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
         }
         
         // Check if signedWeight >= (quorumNumerator/quorumDenominator) * totalWeight
-        // Multiply both sides by quorumDenominator to avoid floating point:
-        // signedWeight * quorumDenominator >= totalWeight * quorumNumerator
-        if (signedWeight * quorumDenominator < totalWeight * quorumNumerator) {
+        // This is equivalent but only using division to avoid any potential overflow
+        if (signedWeight / quorumNumerator < totalWeight / quorumDenominator) {
             revert IWavsServiceManager.InsufficientQuorum();
         }
     }
@@ -259,6 +258,9 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
      *      required for a valid signature (e.g., 2/3 or 51/100)
      */
     function setQuorumThreshold(uint256 numerator, uint256 denominator) external onlyOwner {
+        if (numerator == 0) {
+            revert IWavsServiceManager.InvalidQuorumParameters();
+        }
         if (denominator == 0) {
             revert IWavsServiceManager.InvalidQuorumParameters();
         }
--- a/contracts/eigenlayer/src/WavsServiceManager.sol
+++ b/contracts/eigenlayer/src/WavsServiceManager.sol
@@ -26,8 +26,8 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
     using ECDSAUpgradeable for bytes32;
 
     string public serviceURI;
-    uint256 public quorumNumerator = 2;
-    uint256 public quorumDenominator = 3;
+    uint256 public quorumNumerator;
+    uint256 public quorumDenominator;
 
     constructor(
         address _avsDirectory,
@@ -35,7 +35,7 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
         address _rewardsCoordinator,
         address _delegationManager,
         address _allocationManager
-    )
+    )        
         ECDSAServiceManagerBase(
             _avsDirectory,
             _stakeRegistry,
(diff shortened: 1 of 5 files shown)


ethanfrey:
I did test the code as mentioned in the readme:

```
cd contracts
forge test -C ./eigenlayer -vvv
```

And I ran through the whole docker image manual steps, and it all worked.

This isn't the same as actually testing with multiple WAVS nodes, but I will rely on (and review) the unit tests.

contracts/eigenlayer/src/WavsServiceManager.sol
```diff
@@ -162,22 +168,112 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
...
+        uint256 totalWeight
+    ) internal view {
+        // Avoid 0 weight ever passing this check
+        if (totalWeight == 0) {
+            revert IWavsServiceManager.InsufficientQuorum();
+        }
+        
+        // Check if signedWeight >= (quorumNumerator/quorumDenominator) * totalWeight
+        // This is equivalent but only using division to avoid any potential overflow
+        if (signedWeight / quorumNumerator < totalWeight / quorumDenominator) {
```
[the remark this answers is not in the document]
ethanfrey: It is a trade off between truncation and potential overflow?
Not sure which is more important, but I will go with the current implementation.

contracts/eigenlayer/src/WavsServiceManager.sol
```diff
@@ -162,22 +168,112 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
...
+            );
+        }
+        
+        uint256 totalWeight = registry.getLastCheckpointTotalWeightAtBlock(signatureData.referenceBlock);
+        
+        // Ensure sufficient quorum was reached
+        _validateQuorumSigned(signedWeight, totalWeight);
+    }
+
+    /**
```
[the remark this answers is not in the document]
ethanfrey: Good catch. Some clear cut and pasta. Let's remove the first one

contracts/eigenlayer/src/WavsServiceManager.sol
```diff
@@ -162,22 +168,112 @@ contract WavsServiceManager is ECDSAServiceManagerBase, IWavsServiceManager {
...
+        // Check signature validity
+        if (magicValue != ECDSAStakeRegistry(stakeRegistry).isValidSignature(
+            ethSignedMessageHash,
+            signatureDataBytes
+        )) {
             revert IWavsServiceManager.InvalidSignature();
         }
+
+        // Calculate the total weight of the operators that signed
+        ECDSAStakeRegistry registry = ECDSAStakeRegistry(stakeRegistry);
```
ethanfrey, line 212: This is the actual heart of the code, right? The threshold checks.
There is no cheaper way to do this is there?

It all looks correct. At least matching what I requested

contracts/eigenlayer/test/WavsAVSRegistrar.t.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.9;
+
+import "forge-std/Test.sol";
+import "../src/WavsAVSRegistrar.sol";
+
+contract WavsAVSRegistrarTest is Test {
```
ethanfrey, line 7: Thank you for these follow-up tests on the other review.

contracts/eigenlayer/test/WavsServiceManager.t.sol
```diff
@@ -0,0 +1,272 @@
...
+    
+    function test_validateQuorumSigned_success() public view {
+        // 2/3 of 500 is 333, so 400 should pass
+        serviceManager.validate(
+            IWavsServiceHandler.Envelope({
+                eventId: bytes20(0),
+                ordering: bytes12(0),
+                payload: ""
+            }),
+            createSignatureData(4 * OPERATOR_WEIGHT, 0)
```
ethanfrey: A bit ugly with the OPERATOR_WEIGHT.
Can't you just pass the number of operators that signed and can multiply the OPERATOR_WEIGHT in the function?

contracts/eigenlayer/test/WavsServiceManager.t.sol
```diff
@@ -0,0 +1,272 @@
...
+        );
+        
+        // Test should not revert
+        assertTrue(true, "Validation should pass with sufficient quorum");
+    }
+    
+    function test_validateQuorumSigned_insufficient() public {
```
ethanfrey, line 127: Nice check - yes, this is exactly what we want, 4/5 passes, 3/5 fails.

contracts/eigenlayer/test/WavsServiceManager.t.sol
```diff
@@ -0,0 +1,272 @@
...
+            IWavsServiceHandler.Envelope({
+                eventId: bytes20(0),
+                ordering: bytes12(0),
+                payload: ""
+            }),
+            createSignatureData(3 * OPERATOR_WEIGHT, 0)
+        );
+    }
+    
+    function test_validateQuorumSigned_exact() public {
```
ethanfrey, line 140: Ah, and it updates well.
Nice test coverage

ethanfrey (approved):
Good stuff and thank you for the thorough tests. That is how I like code.

Left a few cleanup comments there. Can you please make those changes, and then I can merge?

ethanfrey:
Thanks for the fixes

PR #125, opened 2025-05-28
feat: `PAST_BLOCKS` optional env

### LINKED ISSUE #123: Make `operator-list` lookback script block lookback configurable
## Summary

ref: https://github.com/Lay3rLabs/wavs-middleware/pull/98#discussion_r2081348458

Turns out that Alchemy has an even more strict limit, 500 blocks :\

Image

## options

1. This should be configurable (BLOCK_LOOKBACK= env)
2. OR we smart query the number of blocks the RPC has and do the request at the limit
3. OR we somehow refactor this so we can query all operators vs it being block lookback based

option 1 seems more reasonable unless eth has an easy way to get this limit / number of blocks available.

reecepbcups:
closes #123 

## Summary

Adds a new (optional) env for the block scanning of OperatorRegister events. Setting PAST_BLOCKS will limit to this amount of lookback. 

### Reason

Some providers allow 2000+ blocks of lookback, others only 1000, and Alchemy only 500. I have not yet tested mainnet but I think they limit it to 64 -128 blocks or something insanely low there. So there is no reasonable set value we can do, just have to make it configurable.

The template will default to 500 for the time being once integrated

### DIFF
--- a/docker/list_operator.sh
+++ b/docker/list_operator.sh
@@ -11,6 +11,8 @@ SCRIPT_DIR="$(realpath "$(dirname "$0")")"
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR"/helpers.sh
 
+PAST_BLOCKS=${PAST_BLOCKS:-1000}
+
 if [ "$DEPLOY_ENV" = "TESTNET" ]; then
     LOCAL_ETHEREUM_RPC_URL="$TESTNET_RPC_URL"
 else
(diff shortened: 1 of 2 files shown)


docker/list_operator.sh
```diff
@@ -11,6 +11,8 @@ SCRIPT_DIR="$(realpath "$(dirname "$0")")"
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR"/helpers.sh
 
+PAST_BLOCKS=${PAST_BLOCKS:-1000}
```
reecepbcups, line 14: Better ENV name ideas? `BLOCK_LOOKBACK` ? or is PAST_BLOCKS good?

ethanfrey: Name sounds good to me.

ethanfrey (approved):
Nice improvement. Thanks

PR #127, opened 2025-05-29
Add getLatestOperatorForSigningKey on service manager

ismellike:
https://github.com/Lay3rLabs/wavs-foundry-template/pull/178#issuecomment-2919396199

Currently getting this error in the template: `error InsufficientQuorum();`

Attempting the suggestion from comment

### DIFF
--- a/contracts/mocks/SimpleServiceManager.sol
+++ b/contracts/mocks/SimpleServiceManager.sol
@@ -95,4 +95,10 @@ contract SimpleServiceManager is IWavsServiceManager {
     function getLastCheckpointTotalWeight() external view returns (uint256) {
         return lastCheckpointTotalWeight;
     }
+
+    function getLatestOperatorForSigningKey(
+        address signingKey
+    ) external pure override returns (address) {
+        return signingKey;
+    }
 }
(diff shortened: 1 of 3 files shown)


contracts/mocks/SimpleServiceManager.sol
```diff
@@ -95,4 +95,10 @@ contract SimpleServiceManager is IWavsServiceManager {
     function getLastCheckpointTotalWeight() external view returns (uint256) {
         return lastCheckpointTotalWeight;
     }
+
+    function getLatestOperatorForSigningKey(
+        address signingKey
+    ) external pure override returns (address) {
+        return signingKey;
```
ethanfrey, line 102: Don’t you want to return operator?
Or this is just a simple mock that returns identity?

ethanfrey (approved):
lgtm. Seems safe to just expose /re-export this

contracts/mocks/SimpleServiceManager.sol
ismellike: Yep just simple mock for wavs tests

PR #132, opened 2025-05-30
Mock emit errors

dakom:
This makes the mock contract emit the same errors as real service manager. Also added a couple more error types so we can be more explicit in tests

### DIFF
--- a/contracts/mocks/SimpleServiceManager.sol
+++ b/contracts/mocks/SimpleServiceManager.sol
@@ -15,11 +15,16 @@ contract SimpleServiceManager is IWavsServiceManager {
         IWavsServiceHandler.Envelope calldata /* envelope */,
         IWavsServiceHandler.SignatureData calldata signatureData
     ) external view override {
-        // Validate that operators are sorted in ascending byte order
-        require(
-            _validateOperatorSorting(signatureData.signers),
-            "Operators are not properly sorted"
-        );
+        // Input validation
+        if (signatureData.signers.length == 0 || signatureData.signers.length != signatureData.signatures.length) {
+            revert IWavsServiceManager.InvalidSignature();
+        }
+        if (signatureData.referenceBlock >= block.number) {
+            revert IWavsServiceManager.InvalidSignature();
+        }
+        if (!_validateOperatorSorting(signatureData.signers)) {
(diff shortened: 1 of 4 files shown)


contracts/mocks/SimpleServiceManager.sol
```diff
@@ -15,23 +15,32 @@ contract SimpleServiceManager is IWavsServiceManager {
...
 
-        // Check if total weight is above threshold
-        require(
-            totalWeight >= lastCheckpointThresholdWeight,
-            "Not enough operator weight"
-        );
+        // Avoid 0 weight ever passing this check
+        if (totalWeight == 0) {
+            revert IWavsServiceManager.InsufficientQuorum();
+        }
```
reecepbcups: Should the error add extra info to say that totalWeight is 0? -- or maybe a new error type IWavsServiceManager.InsufficientQuorumIsZero();

contracts/mocks/SimpleServiceManager.sol
```diff
@@ -15,23 +15,32 @@ contract SimpleServiceManager is IWavsServiceManager {
...
+
+        // Check if the total weight meets the last checkpoint threshold 
+        if (totalWeight < lastCheckpointThresholdWeight) {
+            revert IWavsServiceManager.InsufficientQuorum();
+        }
```
reecepbcups: Same here, showing the weight & checkpoint weight values would probably be useful for debugging?

contracts/mocks/SimpleServiceManager.sol
reecepbcups: i mean these would have to be changed in the parent contract as well **

dakom: done in https://github.com/Lay3rLabs/wavs-middleware/pull/132/commits/c71c9707fc298490b0b5cdee9abd022fbcaa00c7

contracts/mocks/SimpleServiceManager.sol
dakom: done in https://github.com/Lay3rLabs/wavs-middleware/pull/132/commits/49a4feb14537b00eacac732fcd76cd53f39a68ab

reecepbcups (approved):
utAck

PR #133, opened 2025-06-02
Multi cli

ethanfrey:
Improve the cli.sh dispatcher to handle multiple signature formats (for later) and multiple deploy modes (eigenlayer, poa/mirror contract, potentially more).

Also trim down the Dockerfile and help so the scripts can run well locally.

TODO:
- [x] Dispatch over multiple folders
- [x] Move scripts into separate directory (not docker)
- [x] Trim dockerfile size
- [x] Enable scripts to run well locally (outside of docker) if foundry, jq, curl are installed locally.


ethanfrey:
Tested this with all commands in readme, and manually tested some of the routing as well.

Dockerfile down from 476MB to 305MB.

ethanfrey:
Kept on trimming... down to 215MB. Over 50% smaller than it used to be, with all the same functionality.

```
$ docker images | grep wavs-mid
wavs-middleware                                 latest          f73f7a27790c   2 minutes ago    213MB
ghcr.io/lay3rlabs/wavs-middleware               latest          a75d5e017a1d   11 hours ago     489MB
```

reecepbcups (commented):
Tested in template (local only), works as expected :+1:

PR #134, opened 2025-06-02
fix: unbound variable

reecepbcups:
Caught testing against Holesky


reecepbcups:
Going to merge so i can get a tag of it properly vs my local fork

Issue #114, opened 2025-05-21
Create and deploy OpMirrorServiceHandler


ethanfrey:
Depends on #112 and #113 

- Create a new contract in `contracts/poa` called `OpMirrorServiceHandler`. This implements `IServiceHandler` and can call the update method on the `PoaECDSAStakeRegistry` after doing the typical validation on the `ServiceManager`
- Include a deployment script for the `OpMirrorServiceHandler`, including handing over ownership of the `PoaECDSAStakeRegistry`, so it can later update. Code in `contracts/poa/scripts` and `contracts/poa/scripts/utils`
- There should be successful foundry tests showing a deployment of service manager and stake registry (via script lib), followed by deployment of the service handler (via script lib), followed by tests of successful updating of the operator set, as well as failure upon invalid signatures / insufficient quorum.

ethanfrey:
Closed by #136

PR #135, opened 2025-06-06
112 create mirror ecdsastakeregistry

endrit1991:
Fixes issue #112 
Fixes issue #113 

I want to use deployment script for test setUp, so I make one PR for both issues.

Implement MirrorStakeRegistry as subclass of  ECDSAStakeRegistry. Remove normal update methods and add new update method to control validation data by owner. I do not cut and paste the ECDSAStakeRegistry so less code to audit.

PR #136, opened 2025-06-08
114 mirror service handler

endrit1991:
PR comes after #135 
Fixes issue #114 

I make MirrorServiceHandler that takes packet and updates MirrorStakeRegistry.
I add a counter in the packets so they get handled one time in order, and add test cases as you like.
I hope this is what you want. Or must MirrorServiceHandler do more?

### DIFF
--- a/contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
+++ b/contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IWavsServiceHandler} from "../../../interfaces/IWavsServiceHandler.sol";
+import {IWavsServiceManager} from "../../../interfaces/IWavsServiceManager.sol";
+import { MirrorStakeRegistry } from "../MirrorStakeRegistry.sol";
+
+
+interface ITypes {
+    error InvalidTriggerId(uint64 expectedTriggerId);
+
+    /// @notice DataWithId is a struct containing a trigger ID and updated operator info
+    struct UpdateWithId {
+        uint64 triggerId;
+        address[] operators;
+        address[] signingKeys;
+        uint256[] weights;
+    }
+}
+
+
+contract MirrorServiceHandler is ITypes, IWavsServiceHandler {
+    /// @notice Ensures all updates are deployed in order and no duplicates.
+    uint64 public lastTriggerId;
+
+    /// @notice Stake Registry instance
+    MirrorStakeRegistry public stakeRegistry;
+
+    /// @notice Service manager instance
+    IWavsServiceManager public serviceManager;
+
+    constructor(MirrorStakeRegistry _stakeRegistry) {
+        stakeRegistry = _stakeRegistry; 
+        serviceManager = IWavsServiceManager(_stakeRegistry.serviceManager());
+        lastTriggerId = 0;
+    }
+
+    function handleSignedEnvelope(Envelope calldata envelope, SignatureData calldata signatureData) external {
+        // Quick check this is valid trigger id before validating signatures
+        UpdateWithId memory updateData = abi.decode(envelope.payload, (UpdateWithId));
+        uint64 expectedTrigger = lastTriggerId + 1;
+        if (updateData.triggerId != expectedTrigger) {
+            revert InvalidTriggerId(expectedTrigger);
+        }
+
+        // Validate the signatures and update trigger id at this point
+        serviceManager.validate(envelope, signatureData);
+        lastTriggerId = expectedTrigger;
+
+        // call stake registry to update
+        stakeRegistry.batchSetOperatorDetails(updateData.operators, updateData.signingKeys, updateData.weights);
+  }
+
+}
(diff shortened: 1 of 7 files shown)


contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
```diff
@@ -0,0 +1,54 @@
...
+import {IWavsServiceHandler} from "../../../interfaces/IWavsServiceHandler.sol";
+import {IWavsServiceManager} from "../../../interfaces/IWavsServiceManager.sol";
+import { MirrorStakeRegistry } from "../MirrorStakeRegistry.sol";
+
+
+interface IMirrorUpdateTypes {
+    error InvalidTriggerId(uint64 expectedTriggerId);
+
+    /// @notice DataWithId is a struct containing a trigger ID and updated operator info
+    struct UpdateWithId {
```
ethanfrey, line 13: Is this all the items to update that affect `isValidSignature`?

I think there is a theshold weight as well. Can you add this?

contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
```diff
@@ -0,0 +1,54 @@
...
+        if (updateData.triggerId != expectedTrigger) {
+            revert InvalidTriggerId(expectedTrigger);
+        }
+
+        // Validate the signatures and update trigger id at this point
+        serviceManager.validate(envelope, signatureData);
+        lastTriggerId = expectedTrigger;
+
+        // call stake registry to update
+        stakeRegistry.batchSetOperatorDetails(updateData.operators, updateData.signingKeys, updateData.weights);
```
ethanfrey, line 52: yes, this is the basic call to update. good

contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
```diff
@@ -0,0 +1,54 @@
...
+    constructor(MirrorStakeRegistry _stakeRegistry) {
+        stakeRegistry = _stakeRegistry; 
+        serviceManager = IWavsServiceManager(_stakeRegistry.serviceManager());
+        lastTriggerId = 0;
+    }
+
+    function handleSignedEnvelope(Envelope calldata envelope, SignatureData calldata signatureData) external {
+        // Quick check this is valid trigger id before validating signatures
+        IMirrorUpdateTypes.UpdateWithId memory updateData = abi.decode(envelope.payload, (IMirrorUpdateTypes.UpdateWithId));
+        uint64 expectedTrigger = lastTriggerId + 1;
```
ethanfrey: Okay, I like the idea of a triggerId to avoid reply attacks. However, I am unsure if we need to enforce `triggerId = lastTriggerId + 1`. I would propose `tiggerId > lastTriggerId`.

Both of them do not allow reply, or reusing any old unsubmitted operator set.
However, if there is an error somehow reaching consensus on a batch, the `>` type allows to skip some bad one. It also allows us to be more flexible in what we use in WAVS for the triggerId, for example, we could use the block number - which is monotonic and easy to follow. This requires the `>` comparison as well.

contracts/eigenlayer/test/MirrorServiceHandler.t.sol
```diff
@@ -0,0 +1,413 @@
...
+        
+        // Create envelope with the encoded payload
+        IWavsServiceHandler.Envelope memory envelope = IWavsServiceHandler.Envelope({
+            eventId: bytes20(uint160(1)),
+            ordering: bytes12(0),
+            payload: abi.encode(updateData)
+        });
+        
+        // Create signature data with only 3 operators (not enough for quorum)
+        // The quorum is 4/5 (80%) in the default setup
```
ethanfrey, line 209: Huh? ServiceManager quorum is 2/3 right? How do we update this?

Actually, there is no method to change that on the mirror, right? Please add a new ServiceHandler that can be used to sync the changes to quorumNumerator and quorumDenominator in the service manager. You should probably be able to just assign the handler as owner of the service manager to update the fraction needed.

ethanfrey (changes requested):
This is a good start. And thank you for writing some tests on this.

I have 3 major change requests:
1. The new trigger must be greater than the old one, not exactly one more
2. Make sure you sync all the StakeRegistry state that can affect the validation. I think `thresholdWeight` needs to be added, we set this in the shell scripts before. I am not sure if there is anything else. Maybe we sync the service-url also? Or that doesn't matter in the mirrors?
3. Please add a second handler to sync the dynamic threshold of the ServiceManager (default 2/3) to make sure that everything related to verification.

I describe all these points more in detail in the comments.

contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
endrit1991: I updated this

endrit1991:
I made fixes. Now compare trigger is greater.
Added new handler for service manager quorum.
Please review again

contracts/eigenlayer/src/handlers/MirrorServiceHandler.sol
ethanfrey: Please add threshold weight ^^

ethanfrey (commented):
Looks better and 2 of 3 tasks knocked down.
Please add the threshold weight (and test it) and I am good to merge

ethanfrey (approved):
Looks good now.
Thanks for the updates

PR #137, opened 2025-06-11
109 refactor scripts

endrit1991:
Refactor scripts

- [x] cleanup bash
- [x] pause
- [x] unpause
- [x] deploy
- [x] register
- [ ] set_service_uri
- [x] list_operators

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ echo "AVS signing address: $AVS_SIGNING_ADDRESS"
 # Register the operator using the operator key and AVS signing address
 docker run --rm --network host --env-file .env \
    -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
-   wavs-middleware register "$OPERATOR_KEY" "$AVS_SIGNING_ADDRESS" "0.01ether"
+   wavs-middleware register "$OPERATOR_KEY" "$AVS_SIGNING_ADDRESS" "1000000000000000"
 ```
 
 List Operators:
@@ -103,7 +103,6 @@ List Operators:
 # View stake registry status, including registered operators and their weights
 docker run --rm --network host  --env-file .env \
    -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
-   -e PAST_BLOCKS=1000 \
    wavs-middleware list_operator
 ```
 
--- a/contracts/eigenlayer/script/WavsListOperators.s.sol
+++ b/contracts/eigenlayer/script/WavsListOperators.s.sol
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {Script} from "forge-std/Script.sol";
+import {console} from "forge-std/console.sol";
+import {ECDSAStakeRegistry} from "@eigenlayer-middleware/src/unaudited/ECDSAStakeRegistry.sol";
+import {WavsServiceManager} from "../src/WavsServiceManager.sol";
+import {IAllocationManagerTypes, IAllocationManager} from "@eigenlayer/contracts/interfaces/IAllocationManager.sol";
+import {OperatorSet} from "@eigenlayer/contracts/libraries/OperatorSetLib.sol";
+
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+contract WavsListOperators is Script {
+
+    string public constant ENV_SERVICE_MANAGER = "WAVS_SERVICE_MANAGER_ADDRESS";
+
+    struct OperatorInfo {
+        address stakeRegistry;
+        uint256 totalWeight;
+        uint256 thresholdWeight;
+        address[] operators;
+        address[] signingKeys;
+        uint256[] weights;
+    }
+
+    // configuration
+    address private serviceManagerAddr;
+
+    function setUp() public virtual {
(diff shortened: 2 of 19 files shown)


endrit1991:
Ready.
Didn't change set_service-uri as it is one call already

ethanfrey:
I ran through the scripts via readme.md today and it ran. I took a brief look and seems reasonable. Not sure best practices about private keys on foundry. Also, maybe we need some more cleanup removing helpers.sh

but in general looks good

README.md
```diff
@@ -94,7 +94,7 @@ echo "AVS signing address: $AVS_SIGNING_ADDRESS"
 # Register the operator using the operator key and AVS signing address
 docker run --rm --network host --env-file .env \
    -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
-   wavs-middleware register "$OPERATOR_KEY" "$AVS_SIGNING_ADDRESS" "0.01ether"
+   wavs-middleware register "$OPERATOR_KEY" "$AVS_SIGNING_ADDRESS" "1000000000000000"
```
reecepbcups, line 97: I can still do `0.01ether` as the value input upstream right? or did this change to have to use the full wei value?

contracts/eigenlayer/script/WavsListOperators.s.sol
```diff
@@ -0,0 +1,96 @@
...
+        WavsServiceManager serviceManager = WavsServiceManager(serviceManagerAddress);
+        ECDSAStakeRegistry stakeRegistry = ECDSAStakeRegistry(serviceManager.stakeRegistry());
+
+        uint256 totalWeight = stakeRegistry.getLastCheckpointTotalWeight();
+        uint256 thresholdWeight = stakeRegistry.getLastCheckpointThresholdWeight();
+
+        IAllocationManager allocationManager = IAllocationManager(serviceManager.allocationManager());
+        OperatorSet memory opSetQuery = OperatorSet({
+            avs: serviceManagerAddress, 
+            id: 1
```
reecepbcups, line 72: When would id not be 1?

(( this function is so nice! glad we are not scanning events anymore, great refactor)

scripts/ecdsa/eigen/deploy.sh
```diff
@@ -304,34 +66,6 @@ done
 
 cd contracts && forge script eigenlayer/script/WavsMiddlewareDeployer.s.sol --rpc-url $LOCAL_ETHEREUM_RPC_URL --private-key $FUNDED_KEY --broadcast || (echo "Error: Failed to deploy WavsMiddlewareDeployer" && exit 1)
 
-stakeRegistryAddress=$(cat deployments/wavs-middleware/$CHAIN_ID.json | jq -r '.addresses.stakeRegistry')
-WavsServiceManagerAddress=$(cat deployments/wavs-middleware/$CHAIN_ID.json | jq -r '.addresses.WavsServiceManager')
-avsRegistrarAddress=$(cat deployments/wavs-middleware/$CHAIN_ID.json | jq -r '.addresses.avsRegistrar')
-[ -z "$stakeRegistryAddress" -o -z "$WavsServiceManagerAddress" -o -z "$avsRegistrarAddress" ] && { echo "Error: One or more required addresses (stakeRegistryAddress, WavsServiceManagerAddress, avsRegistrarAddress) are null or empty"; exit 1; }
-echo "Middleware contracts deployed with addresses: WavsServiceManager: $WavsServiceManagerAddress, StakeRegistry: $stakeRegistryAddress, AVSRegistrar: $avsRegistrarAddress"
+echo "Middleware contracts deployed with addresses:"
+cat deployments/wavs-middleware/$CHAIN_ID.json | jq .addresses
```
reecepbcups, line 70: nit: `jq .addresses deployments/wavs-middleware/$CHAIN_ID.json` removes the need to pipe

scripts/ecdsa/eigen/pause.sh
```diff
@@ -6,43 +6,21 @@
...
-source "$SCRIPT_DIR"/helpers.sh
-
 if [ "$DEPLOY_ENV" = "TESTNET" ]; then
     LOCAL_ETHEREUM_RPC_URL="$TESTNET_RPC_URL"
 else
     LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL:-http://localhost:8545}
 fi
 
-
+# TODO: this should work outside docker as well - $HOME/.nodes/deployer ?
```
reecepbcups, line 15: in the template we mount to root `-v ./.nodes:/root/.nodes` with docker, so $HOME is not a required change

ethanfrey: $HOME is /root in the docker image.
Not sure the comment, just saying cuz I messed with bash a lot

README.md
endrit1991: It must be full wei. Now foundry script parses env variable: `stakeAmount = vm.envUint(ENV_AMOUNT)`
This parses uint not 0.01ether.

contracts/eigenlayer/script/WavsListOperators.s.sol
endrit1991: We only deploy one operator set per AVS - 1. If we deploy many operator sets, we need more numbers.

created here https://github.com/Lay3rLabs/wavs-middleware/blob/109-refactor-scripts/contracts/eigenlayer/script/utils/WavsMiddlewareDeplomentLib.sol#L107-L119

ethanfrey:
Looks good enough to merge to me. 
Happy for issues for later cleanup @Reecepbcups (like the 0.001ether support, which would be great)

Issue #115, opened 2025-05-21
Bash deploy scripts for PoaStakeRegistry


ethanfrey:
Depends on #114 and #109 

The goal is to take this code and expose it as easy to run cli commands via the docker image. To make that happen, we need to do the following:

- [x] Move the eigenlayer-specific deployment scripts under scripts/eigenlayer
- [x] Make a poa-specific deployment scripts under scripts/poa
- [x] Move cli.sh into scripts
- [ ] Improve the dispatching of `cli.sh` to select script subdir (`--module=poa` or such). Default (no option) is same as `--module=eigenlayer`).
  * This will then dispatch to `scripts/<module>/<command.sh>` with all remaining arguments.
  * You can use getopt, or just make a simple check if the first argument is `--module`
- [ ] Provide useful README documentation showing the new usage
- [ ] Manually test the flow

Note it does not need to deal with multi-chain syncing (that is #110). Just deploying the receiving side here. #110 will build on this

ethanfrey:
closed by #138

PR #138, opened 2025-06-13
115 deploy mirror

endrit1991:
After #137 
Fixes issue #115 

Deploy mirror contracts on a second chain

Now all works. Please review


scripts/ecdsa/mirror/deploy.sh
```diff
@@ -0,0 +1,68 @@
...
+
+if [ -z "$MIRROR_RPC_URL" ]; then
+    echo "Error: MIRROR_RPC_URL is not set in the environment variables (tip: grab from .nodes/avs_deploy.json)."
+    exit 1
+fi
+
+# Set arbitrary location for config
+export WAVS_MIRROR_CONFIG="/wavs/contracts/deployments/wavs-mirror-config.json"
+
+# Read the deployer private key from file
```
ethanfrey, line 32: Do we want the same deployer on the other chain? Or should we pass this in?
I'm not sure here, more a note to self to decide if this is good.

scripts/ecdsa/mirror/deploy.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo "Deployer address: $deployer_address"
+else
+    echo "Error: Deployer key file not found at $HOME/.nodes/deployer"
+    exit 1
+fi
+
+
+if [ "$DEPLOY_ENV" = "LOCAL" ]; then
+    echo "Set gas balancer for deployer on mirror chain"
+    cast rpc anvil_setBalance $deployer_address 0x10000000000000000000 -r $MIRROR_RPC_URL > /dev/null 2>&1 || (echo "Error: Failed to set balance for deployer" && exit 1)
```
ethanfrey, line 46: we probably want some check to ensure balance if this is testnet, I remember the original deploy.sh script before your refactor had something like that.

this code is fine for local, but also want testnet version

scripts/ecdsa/mirror/deploy.sh
```diff
@@ -0,0 +1,68 @@
...
+fi
+
+
+# Read service manager address from file
+export WAVS_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' /root/.nodes/avs_deploy.json)
+
+echo "Reading source chain config:"
+
+cd contracts
+forge script eigenlayer/script/WavsMirrorPrepareDeploy.s.sol --rpc-url $SOURCE_RPC_URL -vv --broadcast
```
ethanfrey: Nice. Two different scripts with different rpc-url. I was wondering how you'd do multichain in solidity.

contracts/eigenlayer/script/utils/WavsMirrorDeploymentLib.sol
```diff
@@ -109,6 +134,72 @@ library WavsMirrorDeploymentLib {
...
+
+        WavsMirrorDeploymentLib.InitialConfiguration memory cfg;
+
+        // get config values
+        cfg.thresholdWeight = stakeRegistry.getLastCheckpointThresholdWeight();
+        cfg.quorumNumerator = serviceManager.quorumNumerator();
+        cfg.quorumDenominator = serviceManager.quorumDenominator();
+
+        // get operators
+        IAllocationManager allocationManager = IAllocationManager(serviceManager.allocationManager());
```
ethanfrey, line 198: okay, looks like the code from list-operators. makes sense

ethanfrey (commented):
Mainly looked at the high-level design in the script.
It seems okay - dumping state from one chain in a json and then loading it to configure the other chain.
I had been wondering how a solidity script could talk to two chains, but this seems like a good approach to just solve it in two steps in bash and not worry about that.

When you get it fully working, let me know and I will run through it and test the docker images and merge it if it works.

scripts/ecdsa/mirror/deploy.sh
ethanfrey: ^^ This is the only real change request in my comments

endrit1991:
I fix comments above. Deploy works now.

ethanfrey (approved):
don't know what changed last commit, but it works.
note: I needed to call `git submodule update --init --recursive` so it would build.

I will merge this, but missing two things I want in a follow up:

1. A script like list operators that works on the mirror, so we can check the state
2. Saving that mirror deployment to a file on the host filesystem. I would say .nodes/mirror-CHAIN_ID.json

Making new follow up issue

PR #140, opened 2025-06-17
139 - Create list_operators script for the mirror version

### LINKED ISSUE #139: PR 138 follow up
This will complete #115 as there were a few details missing, but I didn't want to block merging.

1. Write the output deployment json under /root/.nodes/mirror-$CHAIN_ID.json
2. Create list_operators script for the mirror version.

Supeeerpower:
Fixes #139  

I stored the output deployment json under /root/.nodes/mirror-$CHAIN_ID.json

I created list_operators script for the mirror version.

I stored the list operator output json /root/.nodes/mirror-list-operators-$CHAIN_ID.json.


README.md
```diff
@@ -142,6 +142,15 @@ docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes \
    wavs-middleware -m mirror deploy
 ```
 
+List Mirror Operators:
+
+```bash
+# View stake registry status, including registered operators and their weights
```
reecepbcups, line 156: does this need the WAVS_SERVICE_MANAGER address? how does it know which operators to list on the mirror (i.e. where is the mirror?)

scripts/ecdsa/mirror/list_operators.sh
```diff
@@ -0,0 +1,66 @@
...
+validate_rpc_url "$MIRROR_RPC_URL" "MIRROR_RPC_URL"
+
+# Get chain ID from mirror RPC URL
+MIRROR_CHAIN_ID=$(cast chain-id --rpc-url "$MIRROR_RPC_URL") || error_exit "Failed to get chain ID from mirror RPC URL"
+export MIRROR_CHAIN_ID
+echo "Mirror Chain ID: $MIRROR_CHAIN_ID"
+
+# Read service manager addresses from JSON files
+WAVS_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' /root/.nodes/avs_deploy.json) || error_exit "Failed to read WAVS_SERVICE_MANAGER_ADDRESS from avs_deploy.json"
+MIRROR_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' "/root/.nodes/mirror-$MIRROR_CHAIN_ID.json") || error_exit "Failed to read MIRROR_SERVICE_MANAGER_ADDRESS from mirror-$MIRROR_CHAIN_ID.json"
```
reecepbcups: I don't think we can assume the avs_deploy is on the machine, only the deployer has this. maybe it is fine for just listing, but making a note we moved away from for external interactions for this reason

ref https://github.com/Lay3rLabs/wavs-middleware/pull/98

README.md
Supeeerpower: Ideally we should sync allocation manager as well, but it is a bit complicated.
Current implementation is fetching operator from SOURCE chain, and others from MIRROR chain.

Message from @ethanfrey 

> The most complete might be to sync the allocation manager as well but that is very heavy.
> I think querying the list of operators from the home chain (a separate rpc env var) and then query the rest of the state from the mirror chain.

ethanfrey: agree with Reece here.

* you do need to query both chains.
* you also need both service manager addresses.
* This should not directly read the AVS-deploy file . That is only available on the machine for the dev that deployed, others like operators will just have the variables


It is just moving that logic to parse the AVS-deploy.json away from your script and add them to the readme

scripts/ecdsa/mirror/list_operators.sh
ethanfrey: Same as above.
If you move those lines to the readme that is fine and someone can also set the variables manually

ethanfrey:
One issue is we are making the same scripts for local testing (all on one machine) and real multi node deploys with independent operators.

let’s focus on the second, but maybe add some helper scripts that work for local. You could even make a dir next to scripts/mirror called local or demo that has some commands to run in the local dev environment. Not part of this issue/pr but throwing out the idea for the future, curious what you guys think. It would probably just be a bit of bash calling the same solidity under the hood

README.md
ethanfrey: Add SOURCE_SERVICE_MANAGER and MIRROR_SERVICE_MANAGER

contracts/eigenlayer/script/WavsMirrorListOperators.s.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        // Display results
+        displayResults(opInfo);
+    }
+
+    /// @notice Internal function to fetch operator information from both chains
+    /// @param serviceManagerAddress The address of the source chain service manager
+    /// @param mirrorServiceManagerAddress The address of the mirror chain service manager
+    /// @return OperatorInfo struct containing all operator-related information
+    function listOperators(
```
ethanfrey, line 68: Looks good and nice to see this multi chain aware script

scripts/ecdsa/mirror/deploy.sh
```diff
@@ -8,69 +8,76 @@ export FOUNDRY_DISABLE_NIGHTLY_WARNING=1
 set -o errexit -o nounset -o pipefail
 command -v shellcheck >/dev/null && shellcheck "$0"
 
+# Function to display error message and exit
+error_exit() {
```
ethanfrey, line 12: Nice helper, does make other code cleaner

ethanfrey (commented):
Besides the comments of using env vars rather than the file, this looks good.
Also like the bash cleanup. And some of those can likely go into helpers.sh (which can also be trimmed for stuff we no longer use)

happy to approve with that one change of env vars.

scripts/ecdsa/mirror/list_operators.sh
```diff
@@ -48,19 +48,35 @@ fi
...
 # Get chain ID from mirror RPC URL
 MIRROR_CHAIN_ID=$(cast chain-id --rpc-url "$MIRROR_RPC_URL") || error_exit "Failed to get chain ID from mirror RPC URL"
 export MIRROR_CHAIN_ID
 echo "Mirror Chain ID: $MIRROR_CHAIN_ID"
 
-# Read service manager addresses from JSON files
-WAVS_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' /root/.nodes/avs_deploy.json) || error_exit "Failed to read WAVS_SERVICE_MANAGER_ADDRESS from avs_deploy.json"
-MIRROR_SERVICE_MANAGER_ADDRESS=$(jq -r '.addresses.WavsServiceManager' "/root/.nodes/mirror-$MIRROR_CHAIN_ID.json") || error_exit "Failed to read MIRROR_SERVICE_MANAGER_ADDRESS from mirror-$MIRROR_CHAIN_ID.json"
+# Get service manager addresses from environment variables or files
+SOURCE_SERVICE_MANAGER_ADDRESS=${SOURCE_SERVICE_MANAGER_ADDRESS:-$(jq -r '.addresses.WavsServiceManager' "/root/.nodes/avs_deploy.json")}
```
ethanfrey, line 60: yes good.

docker/env.example
```diff
@@ -15,4 +15,6 @@ FUNDED_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 
 # sETH values for holesky
 LST_CONTRACT_ADDRESS=0x3F1c547b21f65e10480dE3ad8E19fAAC46C95034
-LST_STRATEGY_ADDRESS=0x7D704507b76571a51d9caE8AdDAbBFd0ba0e63d3
\ No newline at end of file
+LST_STRATEGY_ADDRESS=0x7D704507b76571a51d9caE8AdDAbBFd0ba0e63d3
+
+MIRROR_CHAIN_ID=31337
```
ethanfrey, line 20: Ooh, I was working with @Reecepbcups on the mirror chain config - I think he used a different chain-id, it would be good to align with the wavs-bridge one (but I'll leave this for another PR, and Reece might do it himself)

contracts/eigenlayer/script/WavsMirrorListOperators.s.sol
```diff
@@ -98,6 +98,15 @@ contract WavsMirrorListOperators is Script {
             signingKeys[i] = mirrorStakeRegistry.getLatestOperatorSigningKey(operators[i]);
         }
 
+        writeOperatorListJson(block.chainid, OperatorInfo({
```
ethanfrey, line 101: It is fine to write this.
The actual request was to write on MirrorDeploy. Could you add that part in another PR?

contracts/eigenlayer/script/WavsMirrorListOperators.s.sol
```diff
@@ -142,4 +152,42 @@ contract WavsMirrorListOperators is Script {
             console.log(op, sign, weight);
         }
     }
+
+    function writeOperatorListJson(
+        uint256 chainId,
+        OperatorInfo memory opInfo
+    ) internal {
+        string memory fileName = string.concat("deployments/wavs-mirror/list-operators-", vm.toString(chainId), ".json");
```
ethanfrey, line 160: Note, that this is in the docker container, and not on the host file system.
The shell script will typically copy to `/root/.nodes/???` which is mapped to `./.nodes/` so the data persists

ethanfrey (approved):
Good to merge. Some notes for the follow up

contracts/eigenlayer/script/WavsMirrorListOperators.s.sol
Supeeerpower: Could you please elaborate?
Currently, the mirror deploy address is stored in mirror-$CHAIN_ID.json, and the list operators are stored in mirror-list-operators-$CHAIN_ID.json.
image

docker/env.example
Supeeerpower: 31337 is the default chain ID of the anvil fork. If @Reecepbcups  could tell me the mirror chain ID, I could probably configure the anvil fork cli and env.

PR #141, opened 2025-06-17
tmp: output mirror info

reecepbcups:
Used on bridge, needed a temp work around. Will close once middleware saved mirror to file


reecepbcups:
looks like this should be in latest dev, not tested yet but will assume. will re-open if required

PR #148, opened 2025-06-23
131 Mock Deployment Script

### LINKED ISSUE #131: Make mock contract emit same errors as real contract
compare https://github.com/Lay3rLabs/wavs-middleware/blob/cd0ca8699913396c2dcc3fa27a04865af83a1cdb/contracts/mocks/SimpleServiceManager.sol#L33

to

https://github.com/Lay3rLabs/wavs-middleware/blob/cd0ca8699913396c2dcc3fa27a04865af83a1cdb/contracts/eigenlayer/src/WavsServiceManager.sol#L245

Supeeerpower:
Fixes on #131 

- Modify `scripts/ecdsa/mock/deploy.sh` and `contracts/eigenlayer/script/WavsMirrorDeployer.s.sol` for mock deployment.
- Updated `Dockerfile` and `README.md`


docker/mock-config.json
```diff
@@ -0,0 +1,20 @@
+{
```
ethanfrey, line 1: This is nice to have in the repo for reference (point to it in the README), but I wouldn't use it in the image.

I would request people to make their own JSON file and then mount that into the docker image.
Maybe mount it and take the file path as env var to make it super explicit and also to allow it to run outside of docker image

scripts/ecdsa/mirror/deploy.sh
```diff
@@ -37,6 +37,17 @@ MIRROR_CHAIN_ID=$(cast chain-id --rpc-url "$MIRROR_RPC_URL") || error_exit "Fail
 export MIRROR_CHAIN_ID
 echo "Mirror Chain ID: $MIRROR_CHAIN_ID"
 
+# Set deployment stage to MOCK if not set
```
ethanfrey: I would not try to reuse the mirror deploy here.

I would leave the mirror deploy as it and make another one called 'mock/deploy.sh` as a separate file.
This keeps tst code outside of production code and easier to reason about.

You can call that in docker run with `-m mock deploy`
Just make sure all the scripts are copied over in the Dockerfile

contracts/eigenlayer/script/WavsMirrorDeployer.s.sol
```diff
@@ -41,11 +50,11 @@ contract WavsMirrorDeployer is Script, IECDSAStakeRegistryTypes {
         // initialize the operator set
         console2.log("Configuring initial state...");
         WavsMirrorDeploymentLib.setInitialConfiguration(deployment, configuration);
-
-        // deploy the handlers
-        console2.log("Deploying ServiceHandlers as admin...");
-        deployment = WavsMirrorDeploymentLib.deployServiceHandlers(deployment);
-
+        if (deploymentStage == DEPLOYMENT_STAGE_TESTNET) {
```
ethanfrey: Okay, I see the use of reusing the scripts here. But I don't like this variable. I propose two alternate solutions which I am fine with both.

1. Add a boolean env var `WAVS_MIRROR_DEPLOY_HANDLERS` or such and use that to gate the service handler deployment stuff (same logic you have now, just different env variable handling)
2. Copy this deploy script and make a completely different one for mock deploy.

I think the first one might be easier now, but ideally the real deploy becomes a multi-chain aware script later (rather than the intermediate json file) and these two approaches will deviate and no longer fit in one script file.

Dockerfile
```diff
@@ -29,8 +29,10 @@ COPY --from=build --exclude=lib  /wavs/contracts /wavs/contracts/
 # We need this to get the original solidity source from the submodules, but nothing else (15 MB vs 60 MB for all)
 COPY --from=build --parents  /wavs/contracts/./lib/**/*.sol /wavs/contracts/
 
+COPY ./docker/mock-config.json /wavs/docker/mock-config.json
```
ethanfrey: Remove this (see below)

ethanfrey (commented):
I just reviewed commit 5ce19e56f109bce2001610c6ebd940349c50894c

I have not yet reviewed the whole refactoring, so wait until that is approved and merged to try that.

These are comments on the mock deploy part. Good start and just a few adjustments to clean it up.

README.md
```diff
@@ -161,6 +161,58 @@ docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes \
...
+### 3. Deploy the Mock Contracts
+
+```bash
+# Set the path to your local config file
+export LOCAL_CONFIG_PATH=$(pwd)/mock-config.json
+
+# Set the RPC URL for the local blockchain
+export MOCK_RPC_URL=http://localhost:8546
+
+docker run --rm --network host --env-file .env -v ./.nodes:/root/.nodes \
```
ethanfrey, line 215: I'm not sure you need to mount .nodes

Can you try running without mounting, and if it works please remove.
If it breaks, please see if you can remove that dependency (we don't need to use the same deployer key here for the mock)

(This contains info for the source deployment, which doesn't matter here)

scripts/ecdsa/mock/deploy.sh
```diff
@@ -0,0 +1,76 @@
...
+if [ -z "$MOCK_RPC_URL" ]; then
+    error_exit "MOCK_RPC_URL is not set in the environment variables (tip: grab from .nodes/avs_deploy.json)"
+fi
+
+# Get chain ID from rpc url
+MOCK_CHAIN_ID=$(cast chain-id --rpc-url "$MOCK_RPC_URL") || error_exit "Failed to get chain ID from RPC URL"
+export MOCK_CHAIN_ID
+echo "Chain ID: $MOCK_CHAIN_ID"
+
+# Read the deployer private key from file
```
ethanfrey: We don't need this (as this is mainly for the source chain eigenlayer deploy)

I'd just generate a random key and throw it away.

But check with @dakom maybe he wants to modify these mocks later. In which case I'd recommend taking a private key as an env arg (and if missing, just generate one for deploy and don't save it anywhere)

ethanfrey (approved):
Looks good. Minor comment with mounting to use deployer key, which is probably not what wavs needs.

I'm good if you adjust to my suggestions, but always good to double check with @dakom as he'd be the primary user

dakom:
@Supeeerpower - please do feel free to reach out if you want me to pull the latest contracts and run the mocks through WAVS e2e tests, before merging. Might be better on Slack in case we need to iteratively try it out or look at at it together. Happy to do so anytime I'm at the computer, just ping when it's ready and good for you :)

README.md
Supeeerpower: Updated env to use the mock deployer private key.
But I kept to mount .nodes because it stores the mock deploy output.
Please let me know if I don't need to save. then I'll remove the mounting.

scripts/ecdsa/mock/deploy.sh
Supeeerpower: Same reason above, saving mock deploy output in `mock-$MOCK_CHAIN_ID.json`

dakom (approved):
Approving from the perspective of `wavs-tools` using a mock deployment here. The idea of having a predefined JSON file fits perfectly with how we're structured there - i.e. each project may have its own json file, as needed for testing.

Opened https://github.com/Lay3rLabs/wavs-tools/issues/71 to fast-track using this a.s.a.p. so we can give better feedback

My only feedback for now is naming and documentation - what is quorumDenominator and quorumNumerator? How do these relate to threshhold?

Supeeerpower:
> Approving from the perspective of `wavs-tools` using a mock deployment here. The idea of having a predefined JSON file fits perfectly with how we're structured there - i.e. each project may have its own json file, as needed for testing.
> 
> Opened [Lay3rLabs/wavs-tools#71](https://github.com/Lay3rLabs/wavs-tools/issues/71) to fast-track using this a.s.a.p. so we can give better feedback
> 
> My only feedback for now is naming and documentation - what is quorumDenominator and quorumNumerator? How do these relate to threshhold?

Thank you for your reviews @ethanfrey @dakom 

https://github.com/Lay3rLabs/wavs-middleware/blob/dev/contracts/eigenlayer/src/WavsServiceManager.sol#L273
I hope this answers your question.

PR #149, opened 2025-06-24
143 Add Script to Delegate to an Operator

### LINKED ISSUE #143: Add script to delegate to an operator
This can be any funded key, and it should just delegate to an existing operator we previously registered with our scripts.

This can be used in manual testing to change the weights and see how the syncing reacts. See https://github.com/Lay3rLabs/wavs-tools/issues/53

Supeeerpower:
Fixes on #143 

- Added script file to delegate to an operator - `contracts/eigenlayer/script/WavsDelegateToOperator.s.sol` and `scripts/ecdsa/eigen/delegate_to_operator.sh`.
- Update `README.md`.


scripts/ecdsa/eigen/delegate_to_operator.sh
```diff
@@ -0,0 +1,104 @@
...
+    echo "Error: $1" >&2
+    exit 1
+}
+
+SCRIPT_DIR="$(realpath "$(dirname "$0")")"
+# shellcheck source=./helpers.sh
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR"/helpers.sh
+
+if [ -z "${STAKER_KEY:-}" ]; then
```
ethanfrey, line 28: yeah, something like this in the mock deploy contracts

contracts/eigenlayer/script/WavsDelegateToOperator.s.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {Script} from "forge-std/Script.sol";
+import {IDelegationManager} from "@eigenlayer/contracts/interfaces/IDelegationManager.sol";
+import {ISignatureUtilsMixinTypes} from "@eigenlayer/contracts/interfaces/ISignatureUtilsMixin.sol";
+
+import {WavsServiceManager} from "../src/WavsServiceManager.sol";
+
+contract WavsDelegateToOperator is Script {
```
ethanfrey, line 13: I didn't do a deep dive here, but it seems like one step.

When we register an operator, we need to ensure there is eth, then send that to the liquid staking protocol. then deposit that in eigenlayer, then self-delegate to the operator.

I was expected a similar flow here (which could pull a lot of the code from the register_operator. Maybe even refactor that and just call into it to do a few steps.

README.md
```diff
@@ -118,6 +118,27 @@ Unpause Registration:
...
+echo "Staker address: $STAKER_ADDRESS"
+
+export DELEGATION_MANAGER_ADDRESS=0xA44151489861Fe9e3055d95adC98FbD462B948e7
+
+docker run --rm --network host  --env-file .env -v ./.nodes:/root/.nodes \
+   -e STAKER_KEY=${STAKER_KEY} \
+   -e OPERATOR_ADDRESS=${OPERATOR_ADDRESS} \
+   -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
+   -e DELEGATION_MANAGER_ADDRESS=${DELEGATION_MANAGER_ADDRESS} \
+   -e DELEGATION_APPROVER_PRIVATE_KEY=${OPERATOR_KEY} \
```
ethanfrey: This should not require operator private key.

We want to simulate some random staker putting in some sETH on a given operator.

It should need to know the wavs service manager and the operator address to delegate to, not a private key.

I think DELEGATION_MANAGER_ADDRESS can be queried from the WAVS_SERVICE_MANAGER, which owuld make this easier to configure

ethanfrey (commented):
Okay, good start, but some improvements requested

Supeeerpower:
When user registers as an operator, the operator will automatically delegate funds to himself.
However, if user is not an operator, delegating to a registered operator requires the operator's `delegationApprover` permission.
An operator can set the `delegationApprover` address to 0, which means anyone can delegate without permission.
However, the `registerOperator` script sets the `delegationApprover` address to be the same as the operator's own address.
https://github.com/Lay3rLabs/wavs-middleware/blob/dev/contracts/eigenlayer/script/utils/WavsRegisterOperatorLib.sol#L91
(The first parameter is the approver address, if set to 0, the approver private key is not required to delegate.)

The full test should set the approver address separately from the operator address and pass the approver key.
Happy to open a new issue for this or I can add it to another PR.

ethanfrey:
Oh, good catch.

can you modify the registration script to set the delegation approver address to 0? This is the flow we really want

The workflow came from some previous consultants who seemed to have pulled it from hello world. There was no design decision there.

in general, if something doesn’t make sense, it probably wasn’t intentional and good to raise an issue

ethanfrey:
This should let the delegation script be a lot simpler as well. And be much closer to the real deployment we will want

contracts/eigenlayer/script/WavsDelegateToOperator.s.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    error WavsDelegateToOperator__InvalidApproverPrivateKey();
+
+    function setUp() public virtual {
+        wavsServiceManager = WavsServiceManager(vm.envAddress(ENV_SERVICE_MANAGER));
+        delegationManager = IDelegationManager(wavsServiceManager.getDelegationManager());
+        operatorAddress = vm.envAddress(ENV_OPERATOR);
+        delegationApproverAddress = delegationManager.delegationApprover(operatorAddress);
+
+        if (delegationApproverAddress != address(0)) {
```
ethanfrey, line 54: ah, cool, you handle both cases

contracts/eigenlayer/script/WavsDelegateToOperator.s.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    function run() external {
+        vm.startBroadcast();
+
+        ISignatureUtilsMixinTypes.SignatureWithExpiry memory approverSignatureAndExpiry;
+
+        if (delegationApproverAddress == address(0)) {
+            approverSignatureAndExpiry =
+                ISignatureUtilsMixinTypes.SignatureWithExpiry({signature: bytes(""), expiry: 0});
+            delegationManager.delegateTo(operatorAddress, approverSignatureAndExpiry, bytes32(0));
```
ethanfrey, line 74: okay, this still assumes we have some stake we can delegate. I guess adding stake is a different cli call?

(Can you show a workflow in readme that does the full steps?)

scripts/ecdsa/eigen/delegate_to_operator.sh
```diff
@@ -0,0 +1,97 @@
...
+if [ "$DEPLOY_ENV" = "TESTNET" ]; then
+    LOCAL_ETHEREUM_RPC_URL="$TESTNET_RPC_URL"
+    if [ -z "$LOCAL_ETHEREUM_RPC_URL" ]; then
+        echo "Error: TESTNET_RPC_URL environment variable is not set"
+        exit 1
+    fi
+else
+    LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL:-http://localhost:8545}
+    wait_for_ethereum
+    cast rpc anvil_setBalance $staker_address 0x10000000000000000000 -r $LOCAL_ETHEREUM_RPC_URL > /dev/null 2>&1 || (echo "Error: Failed to set balance for deployer" && exit 1)
```
ethanfrey, line 74: okay, good to ensure eth balance, but where do we stake for sETH and stake in eigenlayer?

In register there are some LST_{TOKEN,CONTRACT}_ADDRESS variables to control which LST to use

docker/env.example
```diff
@@ -16,5 +16,3 @@ FUNDED_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 # sETH values for holesky
 LST_CONTRACT_ADDRESS=0x3F1c547b21f65e10480dE3ad8E19fAAC46C95034
```
ethanfrey, line 17: these values

ethanfrey (commented):
This code is fine and can be merged as is, but doesn't close the issue until you actually take eth, stake it in LST and place lst in eigenlayer.

Just reference this code https://github.com/Lay3rLabs/wavs-middleware/blob/dev/contracts/eigenlayer/script/utils/WavsRegisterOperatorLib.sol#L48-L83 (which also checks if there is nothing staked)

Also, you should have an argument for how much eth/shares to stake

contracts/eigenlayer/script/WavsDelegateToOperator.s.sol
```diff
@@ -84,4 +104,54 @@ contract WavsDelegateToOperator is Script {
...
+
+    function _setUpDelegator() internal {
+        // This is the address for private key forge is running the script as.
+        // Calculated from the --private-key argument
+        (, address delegator,) = vm.readCallers();
+
+        IStrategyManager strategyManager = IStrategyManager(delegationManager.strategyManager());
+
+        uint256 numDeposit = strategyManager.stakerStrategyListLength(delegator);
+        if (numDeposit == 0) {
```
ethanfrey, line 116: I think we don’t want this check.
or maybe we do?

one there is a delegation we can delegate again, right? (This was some check for the first time on an operator when it fails halfway through)

contracts/eigenlayer/script/WavsDelegateToOperator.s.sol
```diff
@@ -21,18 +29,28 @@ contract WavsDelegateToOperator is Script {
...
     function setUp() public virtual {
         wavsServiceManager = WavsServiceManager(vm.envAddress(ENV_SERVICE_MANAGER));
-        delegationManager = IDelegationManager(wavsServiceManager.getDelegationManager());
+        delegationManager = DelegationManager(wavsServiceManager.getDelegationManager());
         operatorAddress = vm.envAddress(ENV_OPERATOR);
         delegationApproverAddress = delegationManager.delegationApprover(operatorAddress);
 
+        lstContractAddress = vm.envAddress(ENV_LST_CONTRACT);
+        lstStrategyAddress = vm.envAddress(ENV_LST_STRATEGY);
+        stakeAmount = vm.envUint(ENV_AMOUNT);
```
ethanfrey, line 52: Okay, important to pass this too

scripts/ecdsa/eigen/delegate_to_operator.sh
```diff
@@ -48,6 +48,20 @@ fi
...
+export LST_CONTRACT_ADDRESS
+echo "LST_CONTRACT_ADDRESS: $LST_CONTRACT_ADDRESS"
+
+if [ -z "${LST_STRATEGY_ADDRESS:-}" ]; then
+    error_exit "LST_STRATEGY_ADDRESS is not set in environment variables"
+fi
+export LST_STRATEGY_ADDRESS
+echo "LST_STRATEGY_ADDRESS: $LST_STRATEGY_ADDRESS"
+
+export WAVS_DELEGATE_AMOUNT="$1"
```
ethanfrey, line 63: Nice, cli arg to env for easier solidity usage

ethanfrey (commented):
Please run this script and test it works. That is:

register operator
List operator weight
Delegate more to operator
Check the weight goes up

this can be manual via docker. Better if there is some unit test (but it requires testnet fork, so many not the best for CI)

ensure it works and the same address can delegate multiple times

Once that works (manually tested a few times with the commands in readme) you can merge it.

PR #161, opened 2025-06-30
154 Refactor Shell Script

### LINKED ISSUE #152: Document env vars
The README [suggests to pass the .env file to the docker commands](https://github.com/Lay3rLabs/wavs-middleware/blob/724b19e66de63c3ce4206f0265131697fa3b0848/README.md?plain=1#L64), but the [env file](https://github.com/Lay3rLabs/wavs-middleware/blob/724b19e66de63c3ce4206f0265131697fa3b0848/docker/env.example) contains several vars without explanation (and in some cases pretty confusing, like ETHERSCAN_API_KEY set to a bogus value).

These should be commented with explanation so that users of the middleware can set them with understanding.

ideally, the comments would also say which commands the env vars are used for. Some examples:

* `deploy` needs `FUNDED_KEY`, but `register` does not
* `register` needs `LST_CONTRACT_ADDRESS` but `list-operator` does not

This is helpful since someone might want to use the middleware image for tooling that doesn't follow the precise deploy->register->list flow.

Also, less of a motivation, very much subjective preference, but I prefer calling it like this, so that I can reason about which vars affect which command when debugging deployments:

```
        {{.DOCKER_SUDO}} docker run --rm --network host
        -e FUNDED_KEY 
        -e ETHERSCAN_API_KEY 
        -e DEPLOY_ENV
        -e LOCAL_ETHEREUM_RPC_URL
        -e METADATA_URI 
        -e LST_CONTRACT_ADDRESS 
        -e LST_STRATEGY_ADDRESS
        -e CHAIN_ID 
        -v {{.NODES_DIR}}:/root/.nodes
        {{.MIDDLEWARE_DOCKER_IMAGE}} deplo
```

### LINKED ISSUE #154: Separation of concerns w/ env vars and command args
@ethanfrey feel free to close this if you disagree, it's just a request and felt more correct to open as an issue for now

Currently, some vars are passed via env and some are commandline params. It's especially confusing where `OPERATOR_KEY` is set as an argument, since that's something that might usually be thought of as more fitting for env.

Suggestion:

* All secrets should be in env vars (but only secrets)
* Everything else should be a commandline parameter
* Commandline parameters should be in the form of key=value, allowed in any order
* (optional) Commandline parameters should be able to override env vars

This would follow more typical CLI usage imho

Supeeerpower:
Fixes on #154 #152 

- Update some script to change file config and list operators.
- Refactor shell scripts to ensure structured cli.
- Update Readme


dakom (changes requested):
Sorry for the delayed review, I'm not so familiar with the middleware repo hands-on, so I'm only reviewing from the perspective of using it in `wavs-tools` and the issue I opened at https://github.com/Lay3rLabs/wavs-middleware/issues/154

That said - I love the idea of making reusable functions! Seems like a lot of cleanup was accomplished, and with better logging too. Nicely done!

Marking as "request changes" only because it seems to me that this PR goes the opposite direction of https://github.com/Lay3rLabs/wavs-middleware/issues/154, moving _more_ variables into env vars?

The intent of that issue was to separate env vars (i.e. secrets) from commandline args. So, for example, 


```
export METADATA_URI=https://wavs.xyz/metadata.json
export FUNDED_KEY=
export LST_STRATEGY_ADDRESS=0x7D704507b76571a51d9caE8AdDAbBFd0ba0e63d3

docker run --rm --network host -v ./.nodes:/root/.nodes \
   -e DEPLOY_ENV=${DEPLOY_ENV} \
   -e LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL} \
   -e TESTNET_RPC_URL=${TESTNET_RPC_URL} \
   -e FUNDED_KEY=${FUNDED_KEY} \
   -e METADATA_URI=${METADATA_URI} \
   -e LST_STRATEGY_ADDRESS=${LST_STRATEGY_ADDRESS} \
   wavs-middleware deploy
   ```
 [link](https://github.com/Lay3rLabs/wavs-middleware/pull/161/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R64-R75)
 
 would become:
 
 ```
docker run --rm --network host -v ./.nodes:/root/.nodes \
   -e FUNDED_KEY=${FUNDED_KEY} \
   wavs-middleware deploy \
   DEPLOY_ENV=${DEPLOY_ENV} \
   LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL} \
   TESTNET_RPC_URL=${TESTNET_RPC_URL} \
   METADATA_URI=${METADATA_URI} \
   LST_STRATEGY_ADDRESS=${LST_STRATEGY_ADDRESS} \
 ```
 
 notice how only the `FUNDED_KEY` variable is set as an env var, the rest are commandline parameters.
 
 If this is too hectic to do via the shell script, that's fine, it was just a mild request, not an absolute must
 
 (also, nit, I don't think it's necessary to export vars to the environment _and_ set them via `-e KEY=VALUE`... I could be wrong though, please double-check)

**Comment by Supeeerpower** [comment]
Thank you for great point of view.
Now you can set any variable to env or params, and params will override envs as requirement.
(The funded key can also be set in parameters, which isn't very nice, but I'd like to keep it simple.)
So this is working now.
```
docker run --rm --network host -v ./.nodes:/root/.nodes \
  -e FUNDED_KEY=${FUNDED_KEY} \
  wavs-middleware deploy \
  DEPLOY_ENV=${DEPLOY_ENV} \
  LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL} \
  TESTNET_RPC_URL=${TESTNET_RPC_URL} \
  METADATA_URI=${METADATA_URI} \
  LST_STRATEGY_ADDRESS=${LST_STRATEGY_ADDRESS} \
```

It was a nice catch about `export`, just I'd like to mention the full command here.

**Comment by dakom** [comment]
oh that's excellent, thank you @Supeeerpower !

**Review by dakom — APPROVED** [review]
Untested, but just sorta reading the diff looks good to me. @ismellike maybe you'll already be able to use this in `wavs-tools`. If so, please do try the args syntax (if not, no worries, I'll give it a whirl tomorrow)

File: scripts/helper.sh [hunk 1]
```diff
@@ -0,0 +1,213 @@
+#!/bin/bash
```
ethanfrey, line 1: Cool, nice move to common location

README.md
```diff
@@ -92,30 +111,65 @@ echo "AVS signing address: $AVS_SIGNING_ADDRESS"
...
+export QUORUM_DENOMINATOR=5
+
+docker run --rm --network host -v ./.nodes:/root/.nodes \
+   -e DEPLOY_ENV=${DEPLOY_ENV} \
+   -e LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL} \
+   -e TESTNET_RPC_URL=${TESTNET_RPC_URL} \
+   -e QUORUM_NUMERATOR=${QUORUM_NUMERATOR} \
+   -e QUORUM_DENOMINATOR=${QUORUM_DENOMINATOR} \
+   -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
+   wavs-middleware update_quorum
```
ethanfrey: So, with the parse_args function, you could also call this without the exports above and like:

`docker run ... wavs-middleware update_quorum QUORUM_NUMERATOR=3 QUORUM_DENOMINATOR=5` right?

Can you update the readme to show that usage here, and some other ones that are more user-defined rather than fixed configuration, eg `WAVS_DELEGATE_AMOUNT=1000000000000000` above (but WAVS_SERVICE_MANAGER makes perfect sense as env)

scripts/ecdsa/eigen/update_quorum.sh
```diff
@@ -0,0 +1,41 @@
...
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/../../helper.sh"
+
+# Parse command line arguments in key=value format
+parse_args "$@"
+
+# Check required parameters with defaults
+check_param "DEPLOY_ENV" "${DEPLOY_ENV:-LOCAL}"
+check_param "WAVS_SERVICE_MANAGER_ADDRESS" "${WAVS_SERVICE_MANAGER_ADDRESS:-$(jq -r '.addresses.WavsServiceManager' "$HOME/.nodes/avs_deploy.json")}"
+check_param "QUORUM_NUMERATOR" "${QUORUM_NUMERATOR:-1}"
```
ethanfrey: I don't think `${QUORUM_NUMERATOR:-1}` does what you want. It hardcodes 1 as default, rather than defaulting to `$1`. See my foo.sh example:

```bash
#!/bin/sh

echo ${FOO:-1}
```
Result:
```
$ FOO=3 ./foo.sh
3
$ ./foo.sh
1
$ ./foo.sh 33
1
```

ethanfrey: Try this one:

```bash
#!/bin/sh

echo ${FOO:-$1}
```

Result:

```
$ FOO=3 ./foo.sh
3
$ ./foo.sh 33
33
./foo.sh

```

Which will accept either the env variable set (directly or passed as kv pairs via your parse_args), or the number passed directly as a cli arg. And if neither is passed, it will error.

scripts/ecdsa/eigen/update_quorum.sh
```diff
@@ -0,0 +1,41 @@
...
+# shellcheck source=../../helper.sh
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/../../helper.sh"
+
+# Parse command line arguments in key=value format
+parse_args "$@"
+
+# Check required parameters with defaults
+check_param "DEPLOY_ENV" "${DEPLOY_ENV:-LOCAL}"
+check_param "WAVS_SERVICE_MANAGER_ADDRESS" "${WAVS_SERVICE_MANAGER_ADDRESS:-$(jq -r '.addresses.WavsServiceManager' "$HOME/.nodes/avs_deploy.json")}"
```
ethanfrey: This is a rather complex line. I would split it into two lines for legibiltiy:

```bash
DEFAULT_SERVICE_MANAGER=$(jq -r '.addresses.WavsServiceManager' "$HOME/.nodes/avs_deploy.json" || true)
check_param "WAVS_SERVICE_MANAGER_ADDRESS" "${WAVS_SERVICE_MANAGER_ADDRESS:-$DEFAULT_SERVICE_MANAGER}"
```

Also, the `|| true` means if the .node file is not readable, it will give an empty default. But if `WAVS_SERVICE_MANAGER_ADDRESS` is set explicitly, this will not break the script. Allowing either one.

scripts/helper.sh
```diff
@@ -0,0 +1,213 @@
...
+
+# Check if a required parameter is set (either from env or command line)
+# Usage: check_param "PARAM_NAME" "$PARAM_VALUE"
+check_param() {
+    local param_name="$1"
+    local param_value="$2"
+    if [ -z "$param_value" ]; then
+        handle_error "$param_name is not set (check environment variables or command line parameters)"
+    else
+        export "$param_name"="$param_value"
```
ethanfrey, line 44: like this script. and then explicitly setting this if it read the value from some default (`${DEPLOY:-LOCAL}`) is a great idea

scripts/helper.sh
```diff
@@ -0,0 +1,213 @@
...
+}
+
+# Set up environment based on DEPLOY_ENV
+# Usage: setup_environment
+setup_environment() { 
+    # Set up environment based on DEPLOY_ENV
+    if [ "$DEPLOY_ENV" = "LOCAL" ]; then
+        check_param "LOCAL_ETHEREUM_RPC_URL" "${LOCAL_ETHEREUM_RPC_URL:-http://localhost:8545}"
+    else
+        check_param "TESTNET_RPC_URL" "${TESTNET_RPC_URL:-}"
```
ethanfrey, line 121: why empty default here?

`${TESTNET_RPC_URL:-}` rather than `${TESTNET_RPC_URL}`

(You may have a good reason, it was just not obvious to me what the difference was, happy to hear the explanation)

scripts/helper.sh
```diff
@@ -0,0 +1,213 @@
...
+    export LOCAL_ETHEREUM_RPC_URL
+    echo "Using RPC URL: $LOCAL_ETHEREUM_RPC_URL"
+
+    # Wait for Ethereum node to be ready
+    wait_for_ethereum
+}
+
+# Ensure an account has sufficient balance
+# Usage: ensure_balance "0x..." [rpc_url]
+ensure_balance() {
```
ethanfrey, line 133: i like this one. combined both approaches nicely (for testnet and local dev)

scripts/helper.sh
```diff
@@ -0,0 +1,213 @@
...
+show_usage() {
+    local script_name="$1"
+    local description="$2"
+    local usage_example="$3"
+    
+    echo "Usage: $script_name"
+    echo "Description: $description"
+    echo "Example: $usage_example"
+    echo ""
+    echo "Environment variables (secrets):"
```
ethanfrey: Are these the same for all scripts?
Is this even used?

It looks good for the deploy script, but then I'd move it into that script alone, not as a helper (as this seems script-specific). In general, happy for each script to have it's show_usage if you find that useful. Just don't see this as a generic helper

scripts/helper.sh
```diff
@@ -0,0 +1,213 @@
...
+    do
+        echo "Waiting for $LOCAL_ETHEREUM_RPC_URL"
+        sleep 1
+    done
+    echo "Ethereum node is ready!"
+}
+
+# Impersonate an account (for local development)
+# Usage: impersonate_account "0x..."
+impersonate_account() {
```
ethanfrey, line 89: Are these still used? There are a few functions here which seem left over from old (pre-refactor) code, and happy for you to remove any functions that are not referenced by current scripts.

ethanfrey (commented):
In general, I love this refactor. One helpers.sh location, and all the scripts are so much cleaner to read now. And there is even more flexibility and power.

I added a number of details / cleanup questions. Mainly focusing on the update_quorum script I just reviewed.  With this cleanup, happy to merge.

Main functional questions (rather than style questions) are:
* Supporting cli args (eg "update_quorum 3 5") in update_qourum
* Not using 1/2 default

If you want to promote the use of parse_args, it is great to show the use of these key=value pairs in some of the readme examples where it makes sense (I pointed out one or two)

If you can do a bit of cleanup here, I will re-review and then happy to merge

scripts/helper.sh
Supeeerpower: If you don't mind, I'll leave this part as is for now, as it's used in both ecdsa and bls's `set_service_uri.sh`. In the future, I'll add a Solidity script to set the service URI, and I'll also update `helper.sh`.

scripts/helper.sh
Supeeerpower: Here is my thought, If we pass empty default value, then it will revert with our custom error here.
```
check_param() {
    local param_name="$1"
    local param_value="$2"
    if [ -z "$param_value" ]; then
        handle_error "$param_name is not set (check environment variables or command line parameters)"
    else
        export "$param_name"="$param_value"
        echo "Using value for $param_name: $param_value"
    fi
}
```
Just my thought, happy to remove if you want.

README.md
ethanfrey: fixed

scripts/ecdsa/eigen/update_quorum.sh
ethanfrey: fixed

scripts/helper.sh
ethanfrey: And if we don't provide empty default, what happens?

scripts/helper.sh
ethanfrey: fair enough. no nedd to remove this or anything else, just thought it was left over.

ethanfrey (approved):
Looks good. Let's merge

PR #170, opened 2025-07-08
168 Revert when No Operator in Mirror Deployment

### LINKED ISSUE #168: When deploy ecdsa mirror, ensure there are operators
This is a footgun that hit a few people.

They deployed the eigenlayer contracts to the source chain.
Then they deployed the mirror contracts to the second chain.
Then they registered the operators.

However, the mirror contract is set with no operators at all and thus will never approve any message from the service handlers in order to sync it later.

We should be better and fail faster so people spot this error earlier. Update https://github.com/Lay3rLabs/wavs-middleware/blob/dev/contracts/eigenlayer/script/utils/WavsMirrorDeploymentLib.sol such that when setting the initial configuration on the mirror chain, it will revert if there are no operators there.

Supeeerpower:
Fixes on #168 

Revert when no operator in mirror deployment.


ethanfrey (approved):
Looks good to me. I would love to see a testcase for this, but if it is too much work to add the first one for this deployer script, then it can wait.

If you can make a simple test for that, great. If you can't do that in 45 minutes, just merge as is.

PR #172, opened 2025-07-09
169 Script to Deregister Operator

### LINKED ISSUE #169: Script to deregister operator
Useful for https://github.com/Lay3rLabs/wavs-tools/issues/54

We have scripts for registering an operator and delegating to an operator, but we don't have anything to deregister an operator as far as I can tell.

With this I can easily trigger my component and create a test-case.

Supeeerpower:
Fixes on #169 

- Add deregister solidity and shell script
- Update README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,18 @@ docker run --rm --network host \
    wavs-middleware register WAVS_DELEGATE_AMOUNT=1000000000000000
 ```
 
+Deregister:
+
+```bash
+docker run --rm --network host \
+   -e DEPLOY_ENV=${DEPLOY_ENV} \
+   -e LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL} \
+   -e TESTNET_RPC_URL=${TESTNET_RPC_URL} \
+   -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
+   -e OPERATOR_KEY=${OPERATOR_KEY} \
+   wavs-middleware deregister
+```
+
 List Operators:
 
 ```bash
--- a/contracts/script/eigenlayer/ecdsa/utils/WavsRegisterOperatorLib.sol
+++ b/contracts/script/eigenlayer/ecdsa/utils/WavsRegisterOperatorLib.sol
@@ -168,4 +168,45 @@ library WavsRegisterOperatorLib {
             console2.log("Operator %s is already registered with AVS", operatorAddr);
         }
     }
+
+    function deregisterFromAvs(
+        address serviceManagerAddress
+    ) internal {
+        WavsServiceManager serviceManager = WavsServiceManager(serviceManagerAddress);
+        ECDSAStakeRegistry stakeRegistry = ECDSAStakeRegistry(serviceManager.stakeRegistry());
+
+        // This is the address for private key forge is running the script as.
+        // Calculated from the --private-key argument
+        (, address operatorAddr,) = VM.readCallers();
+
+        // query if already in opset and add if if not in it yet.
+        IAllocationManager allocationManager =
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -118,6 +118,18 @@ docker run --rm --network host \
...
 
+Deregister:
+
+```bash
+docker run --rm --network host \
+   -e DEPLOY_ENV=${DEPLOY_ENV} \
+   -e LOCAL_ETHEREUM_RPC_URL=${LOCAL_ETHEREUM_RPC_URL} \
+   -e TESTNET_RPC_URL=${TESTNET_RPC_URL} \
+   -e WAVS_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
+   -e OPERATOR_KEY=${OPERATOR_KEY} \
```
ethanfrey, line 129: Looks good, just needs the operator key

contracts/script/eigenlayer/ecdsa/utils/WavsRegisterOperatorLib.sol
```diff
@@ -168,4 +168,45 @@ library WavsRegisterOperatorLib {
...
+                avs: serviceManagerAddress,
+                operatorSetIds: opSetIds
+            });
+            allocationManager.deregisterFromOperatorSets(params);
+
+            console2.log(
+                "Successfully deregistered operator %s from operator sets [1]", operatorAddr
+            );
+        } else {
+            console2.log("%s not registered to operator sets [1]", operatorAddr);
```
ethanfrey, line 201: We should return early here, right?

I would make the if statement check if the operator is NOT in the operator set, and then print this console log and return early. `allocationManager.deregisterFromOperatorSets` would be in the main flow (not inside an if statement)

ethanfrey: Nevermind, maybe the registered but are not in any operator sets.

Your code flow is correct.

contracts/script/eigenlayer/ecdsa/utils/WavsRegisterOperatorLib.sol
```diff
@@ -168,4 +168,45 @@ library WavsRegisterOperatorLib {
...
+        }
+
+        if (stakeRegistry.operatorRegistered(operatorAddr)) {
+            console2.log("Deregistering operator %s with AVS ...", operatorAddr);
+            stakeRegistry.deregisterOperator();
```
ethanfrey, line 206: We need to deregister from the stake manager separately from the allocation manager? One doesn't call the other?

Not a problem if this works, I just  figured they'd have some hooks :shrug:

scripts/ecdsa/eigen/deregister.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
ethanfrey, line 1: I like this script. So easy to read now with your bash refactor

ethanfrey (approved):
Good stuff. I like the code.

Was making a comment to change the flow, but after reflection, I think you are right.

I did not test this. If @ismellike could run this locally, and approve before merge, that would be great

contracts/script/eigenlayer/ecdsa/utils/WavsRegisterOperatorLib.sol
Supeeerpower: In BLS stuff, it works like you said.

Here in ECDSA, `deregisterFromOperatorSets` is calling `AVSRegistrar.deregisterOperator` which is mock right now.
https://github.com/Lay3rLabs/wavs-middleware/blob/169-script-to-deregister-operator/contracts/src/eigenlayer/ecdsa/WavsAVSRegistrar.sol#L42

PR #185, opened 2025-07-10
183 Separated Interface for ServiceHandler

### LINKED ISSUE #183: Move UpdateWithId to its own file
A common pattern we've used is to move types that should be used by components to their own file (types.sol or similar up to you)

Wavs-tools needs to use this UpdateWithId type, but it's in the ServiceHandler atm. Trying to sol! macro the whole file gives a rust error, because types are colliding with other sol macros I'm using.
https://github.com/Lay3rLabs/wavs-middleware/blob/0ffb0abf61d284bef54e26617a1009a290bfe3ad/contracts/src/eigenlayer/ecdsa/handlers/MirrorServiceHandler.sol#L8-L19

Same with this: https://github.com/Lay3rLabs/wavs-middleware/blob/0ffb0abf61d284bef54e26617a1009a290bfe3ad/contracts/src/eigenlayer/ecdsa/handlers/MirrorServiceManagerHandler.sol#L11-L15 for next PR

Supeeerpower:
Fixes on #183 

Add separated interface files for service handler contracts

### DIFF
--- a/contracts/src/eigenlayer/ecdsa/interfaces/IMirrorServiceHandler.sol
+++ b/contracts/src/eigenlayer/ecdsa/interfaces/IMirrorServiceHandler.sol
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+interface IMirrorUpdateTypes {
+    error InvalidTriggerId(uint64 expectedTriggerId);
+
+    /// @notice DataWithId is a struct containing a trigger ID and updated operator info
+    struct UpdateWithId {
+        uint64 triggerId;
+        uint256 thresholdWeight;
+        address[] operators;
+        address[] signingKeyAddresses;
+        uint256[] weights;
+    }
+}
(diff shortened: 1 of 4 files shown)


ismellike (commented):
I find it a bit weird that the file name is different from the contents.
Would it make sense to just name IMirrorUpdateTypes.sol?

Supeeerpower:
> I find it a bit weird that the file name is different from the contents. Would it make sense to just name IMirrorUpdateTypes.sol?

If you have any specific requirements regarding this, please let me know and I will make modifications for you, but I would like to keep the interface names that are commonly used in Solidity. I think it would be possible to change the interface file structure as more features are added. I think it's ok for now.
```
IMock.sol
├── IMockTypes
├── IMockErrors
├── IMockEvents is IMockTypes
└── IMock is IMockErrors, IMockEvents
Mock.sol
└── Mock is IMock
```

contracts/src/eigenlayer/ecdsa/interfaces/IMirrorServiceHandler.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+interface IMirrorUpdateTypes {
```
ethanfrey, line 4: Maybe IStakeUpdateTypes?

Both this and the other ones are for the mirror

ethanfrey (approved):
Looks good to me with possible naming change.

@ismellike should check this makes the Rust compiler happier. Probably best to wait for that confirmation, but not required.

ismellike:
lgtm

PR #189, opened 2025-07-14
181 Cleanup Script for Multiple Deployment

### LINKED ISSUE #181: Better support for multiple deployments
Right now, everytime we deploy, we overwrite the `docker/.nodes` dir. This contains the avs_deploy.json (of the deployed contracts) and the `deployer` private key. Which is great info to have to update the service, register on it, etc.

However, this should be extended to support deploying multiple services on one chain, as well as deploying on multiple chains. I am open to comments on how to better organize this, but my thoughts are:

- [ ] include the chain_id of the source chain in `.nodes/avs_deploy.json`
- [ ] naming conventions and documentation on how to manage multiple deployments (we could map different directories to `/root/.nodes` for each deploy and it will work, but some tooling will make it work better)
- [ ] tooling to help manage multiple env files, one per chain_id (or deployment? maybe after we deploy, we can copy the .env file into the deployment-specific directory)

Some rough idea would be like this:

```bash
cd docker
cp .env.example.holesky .env
# TODO: modify any variables there...

# Deploy
docker run --rm --network host -v ~/.wavs/hole-art:/root/.nodes \
   --env-file .env \
   wavs-middleware deploy

# Store the env file
cp .env ~/.wavs/hole-art

# Run future commands in context of the specific deployment
cd ~/.waves/hole-art
docker run --rm --network host -v .:/root/.nodes \
   --env-file .env \
   wavs-middleware set_service_uri SERVICE_URI="https://ipfs.url/for-custom-service.json"
```

We could also add some helpers here to list deployments or change to them, eg:

```bash
alias list-deploys="ls ~/.wavs"
# Note: this should have error checking that $1 is filled and a usage message if not, but fundamentally this works 
use-deploy() { cd ~/.wavs/$1; }
```

Supeeerpower:
Fixes on #181 

- Updated file path for deploy, list operators.
- Passed deployer private key as params option.
- Removed chain_id in output files.

### DIFF
--- a/scripts/bls/eigen/pause.sh
+++ b/scripts/bls/eigen/pause.sh
@@ -23,7 +23,7 @@ check_param "REGISTRY_ADDRESS" "${REGISTRY_ADDRESS:-$DEFAULT_REGISTRY_ADDRESS}"
 setup_environment
 
 # Read the deployer private key
-deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer" || true)
+deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer")
 check_param "FUNDED_KEY" "${FUNDED_KEY:-$deployer_private_key}"
 deployer_address=$(cast wallet address "$FUNDED_KEY")
 echo "Deployer address: $deployer_address"
--- a/scripts/bls/eigen/pause.sh
+++ b/scripts/bls/eigen/pause.sh
@@ -16,14 +16,16 @@ parse_args "$@"
 
 # Check required parameters with defaults
 check_param "DEPLOY_ENV" "${DEPLOY_ENV:-LOCAL}"
-check_param "SLASHING_REGISTRY_COORDINATOR_ADDRESS" "${SLASHING_REGISTRY_COORDINATOR_ADDRESS:-$(jq -r '.addresses.registryCoordinator' "$HOME/.nodes/avs_deploy.json")}"
+DEFAULT_REGISTRY_ADDRESS=$(jq -r '.addresses.registryCoordinator' "$HOME/.nodes/avs_deploy.json" || true)
+check_param "REGISTRY_ADDRESS" "${REGISTRY_ADDRESS:-$DEFAULT_REGISTRY_ADDRESS}"
 
 # Set up environment based on DEPLOY_ENV
 setup_environment
 
-# Read the deployer private key from file
-deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer")
-deployer_address=$(cast wallet address "$deployer_private_key")
+# Read the deployer private key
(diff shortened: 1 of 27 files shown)


scripts/bls/eigen/pause.sh
```diff
@@ -16,14 +16,16 @@ parse_args "$@"
...
 
 # Set up environment based on DEPLOY_ENV
 setup_environment
 
-# Read the deployer private key from file
-deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer")
-deployer_address=$(cast wallet address "$deployer_private_key")
+# Read the deployer private key
+deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer" || true)
+check_param "FUNDED_KEY" "${FUNDED_KEY:-$deployer_private_key}"
```
ismellike, line 27: Would it make more sense for this arg to just be DEPLOYER_PK instead of FUNDED_KEY? (and throughout other scripts)

ismellike (commented):
Thanks for looking through this. From a quick scan, it looks like the issues were addressed, so I'll work on integrating this with wavs-tools and confirm. I just left 1 question below.

scripts/bls/eigen/pause.sh
reecepbcups: FUNDED_KEY is likely left over from Breadchain's middleware work and is used heavily in the template and our current examples. If we do rename, this would be a breaking change

scripts/ecdsa/mirror/deploy.sh
```diff
@@ -26,14 +26,10 @@ else
...
-# Get chain ID from mirror RPC URL
-MIRROR_CHAIN_ID=$(get_chain_id "$MIRROR_RPC_URL")
-export MIRROR_CHAIN_ID
-echo "Mirror Chain ID: $MIRROR_CHAIN_ID"
-
-# Read the deployer private key from file
-deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer")
-deployer_address=$(cast wallet address "$deployer_private_key")
+# Read the deployer private key
+deployer_private_key=$(load_deployment_data "$HOME/.nodes/deployer" || true)
```
ismellike: I get this error here even though I pass in FUNDED_KEY

```
Calling: /wavs/scripts/ecdsa/mirror/deploy.sh
Using value for DEPLOY_ENV: LOCAL
Using value for SOURCE_RPC_URL: http://localhost:8548
Using value for MIRROR_RPC_URL: http://localhost:8549
Error: Deployment file /root/.nodes/deployer not found
```

Supeeerpower: Sorry I didn't test it myself, I just had too many other things to care about in this pr.

ismellike (approved):
lgtm wrt issues - got to this pretty late in day but can confirm working ty
utack on other stuff like renaming

PR #190, opened 2025-07-14
92 Multi Deployment and env.example (sepolia)

### LINKED ISSUE #92: Migrate to Sepolia
Eigen contracts found here

https://github.com/Layr-Labs/eigenlayer-contracts-zeus-metadata/blob/master/environment/testnet-sepolia/manifest.json

Supeeerpower:
Fixes on #92 

- Updated holesky addresses.
- Add hoodi and sepolia configurations.
- Add the method to use multiple env.example, updated README.md

### DIFF
--- a/contracts/deployments/eigenlayer-core/11155111.json
+++ b/contracts/deployments/eigenlayer-core/11155111.json
@@ -1,0 +1,25 @@
+{
+  "lastUpdate": { "timestamp": "1752504238", "block_number": "0" },
+  "addresses": {
+    "proxyAdmin": "0x56E88cb4f0136fC27D95499dE4BE2acf47946Fa1",
+    "delegation": "0xD4A7E1Bd8015057293f0D0A557088c286942e84b",
+    "delegationManagerImpl": "0xa9821F0620D347648f375c597761C7FD16C1b823",
+    "avsDirectory": "0xa789c91ECDdae96865913130B786140Ee17aF545",
+    "avsDirectoryImpl": "0xD88b96998325c3e74A74a0B0938BBFeA1395C188",
+    "strategyManager": "0x2E3D6c0744b10eb0A4e6F679F71554a39Ec47a5D",
+    "strategyManagerImpl": "0xf0A8735c26121e6C488ebac65c8fa3fe37cBFCB3",
+    "eigenPodManager": "0x56BfEb94879F4543E756d26103976c567256034a",
+    "eigenPodManagerImpl": "0x10848d61cE044bcAEE301a28A4C30Ad0a40e88Ae",
+    "strategyFactory": "0x066cF95c1bf0927124DFB8B02B401bc23A79730D",
+    "strategyFactoryImpl": "0xEE41826B7D5B89e7F5eED6a831b4eFD69FC9ca1A",
+    "strategyBeacon": "0x427e627Bc7E83cac0f84337d3Ad94230C32697D3",
+    "strategyBeaconImpl": "0x3BA4F6973D9ed2f0617C4b15d8CE06155fd6a43b",
+    "allocationManager": "0x42583067658071247ec8CE0A516A58f682002d07",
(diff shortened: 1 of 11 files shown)


contracts/deployments/eigenlayer-core/17000.json
```diff
@@ -1,23 +1,25 @@
 {
-  "lastUpdate": { "timestamp": "1733242248", "block_number": "0" },
+  "lastUpdate": { "timestamp": "1752504238", "block_number": "0" },
```
ethanfrey, line 2: What is with these updates on holesky?

Oh, it seems that the delegation address is the same but delegationImpl changed? So they did a proxy upgrade since we made this file, right? But the main addresses are the same

contracts/deployments/eigenlayer-core/11155111.json
```diff
@@ -0,0 +1,25 @@
...
+    "strategyBeacon": "0x427e627Bc7E83cac0f84337d3Ad94230C32697D3",
+    "strategyBeaconImpl": "0x3BA4F6973D9ed2f0617C4b15d8CE06155fd6a43b",
+    "allocationManager": "0x42583067658071247ec8CE0A516A58f682002d07",
+    "allocationManagerImpl": "0xB87AeeA46BB3a3050D277272E33b011922537Db7",
+    "rewardsCoordinator": "0x5ae8152fb88c26ff9ca5C014c94fca3c68029349",
+    "rewardsCoordinatorImpl": "0xcC305562B01bec562D13A40ef8781e313AFE7940",
+    "permissionController": "0x44632dfBdCb6D3E21EF613B0ca8A6A0c618F5a37",
+    "permissionControllerImpl": "0x59B11b191B572888703E150E45F5015e0fFcf525"
+  },
+  "metaDataURI": "https://raw.githubusercontent.com/ethgas-developer/ethgas-developer.github.io/main/vision-avs-2.json"
```
ethanfrey: We should probably have a different default metadataURI for all of these, not even sure if that is used.

Not blocking this PR at all, just wanted to bring it up - I'm not sure how we handle that and if we have some default URL for all wavs apps.

contracts/deployments/eigenlayer-core/11155111.json
```diff
@@ -0,0 +1,25 @@
+{
```
ethanfrey, line 1: Note: I don't know what chain this is. You can't add comments to json, but it would be good to include some docs on which chain id is what common name / rpc endpoint / etc

contracts/deployments/strategies/11155111.json
```diff
@@ -0,0 +1,7 @@
+{
+  "strategies": [
+    "0x424246eF71b01ee33aA33aC590fd9a0855F5eFbc",
```
ethanfrey, line 3: Again, it would be good to document which ones these are. I guess sETH.

Not sure where

docker/env.example.sepolia
```diff
@@ -0,0 +1,17 @@
+# source rpc url for local deployment
+FORK_RPC_URL=https://ethereum-sepolia-rpc.publicnode.com
+
+DEPLOY_ENV=LOCAL
+ETHERSCAN_API_KEY=
+LOCAL_ETHEREUM_RPC_URL=http://localhost:8545
+TESTNET_RPC_URL=
+
+# FIXME: place your project's metadata URI here
+METADATA_URI=https://wavs.xyz/metadata.json
```
ethanfrey, line 10: Yeah, this is the one I meant about metadata uri. There is nothing on this link.
See #200 for the follow up

ethanfrey (approved):
Looks good. I did not try it, but the changes make sense.

The main change request is adding some documentation in the README

contracts/deployments/eigenlayer-core/17000.json
Supeeerpower: Yeah, eigenlayer updated their implementations, but it doesn't really matter for us.
Actually we don't even need implementation address.

PR #194, opened 2025-07-16
193 CLI Update to Support BLS and ECDSA

### LINKED ISSUE #193: CLI Update to Support BLS and ECDSA
@ismellike raised the issue.

Some contracts with the same name in the BLS and ECDSA build outputs are overwritten and can only be used in one output.

- Need to separated output directory.
- Update Dockerfile, shell script and Makefile to support this feature.
- Added some view functions based on request from @ismellike

Supeeerpower:
Fixes on #193 

- Separate build output directory of bls and ecdsa.
- Update Dockerfile, shell script and Makefile to support this feature.
- Added some view functions based on request from @ismellike


ismellike (approved):
lgtm - in terms of things I need from IWavsServiceManager
personally not sure about BLS side will defer to you for that

I dont think the foundry profile is useful for me though. These are all commands you run within the repo. All good if it works for you and makes sense.

Supeeerpower:
You can now build all contracts for bls and ecdsa using the `make build` CLI.
The bls build output is in `out/bls`, and the ecdsa build output is in `out/ecdsa`.

PR #196, opened 2025-07-17
Add SimpleServiceManager Test

### LINKED ISSUE #153: Add test coverage to mock contracts
There is test coverage for the real service manager: https://github.com/Lay3rLabs/wavs-middleware/blob/53800eade09a9747bd0bec6157f3c6772d19a839/contracts/test/eigenlayer/ecdsa/WavsServiceManager.t.sol

However, there is not test coverage for the mock contracts. Ideally this would be abstracted so that the same test suite runs over both contracts, to make sure mocks are covering the same API as the real contracts, but any test coverage here is good :)

Supeeerpower:
Fixes on #153 

- Add test for SimpleServiceManager.sol.
- Fixed some warning from Solhint.

### DIFF
--- a/.solhint.json
+++ b/.solhint.json
@@ -8,6 +8,8 @@
     "no-inline-assembly": "off",
     "no-console": "off",
     "import-path-check": "off",
-    "no-empty-blocks": "off"
+    "no-empty-blocks": "off",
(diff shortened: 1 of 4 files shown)


dakom (approved):
LGTM!

PR #199, opened 2025-07-18
test(e2e): docci integration

reecepbcups:
Builds on top of https://github.com/Lay3rLabs/wavs-middleware/pull/198

## Summary

Ethan mentioned before Berlin that it was a pain for him to have to manually run the markdown to verify changes.

This integrates [Docci](https://github.com/Reecepbcups/docci), a tool I wrote that turns your markdown files into testsuites without writing any code (just these markdown tags), just context. I also use this in the [wavs-bridge](https://github.com/Lay3rLabs/wavs-bridge/pull/56) as well.

It will catch any commands with non 0 exit status (perfect for `set -e` scripts), or if the stdout/err does not contain the desired text. useful to have general sanity checks along the way. including echo'ing a variable and making sure its set.

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -22,7 +22,7 @@ docker build -t wavs-middleware .
 
 start holesky fork:
 
-```bash
+```bash docci-background docci-delay-after=2
 RPC_URL=https://ethereum-holesky-rpc.publicnode.com
 anvil --fork-url $RPC_URL --host 0.0.0.0 --port 8545
 ```
@@ -73,7 +73,7 @@ docker run --rm --network host -v ./.nodes:/root/.nodes \
 
 list operators:
 
-```bash
+```bash docci-output-contains="Operator 1:"
 docker run --rm --network host -v ./.nodes:/root/.nodes \
   wavs-middleware list_operators
 ```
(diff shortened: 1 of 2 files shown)


DEV.md
```diff
@@ -139,14 +139,23 @@ docker run --rm --network host -v ./.nodes:/root/.nodes \
...
+MIRROR_RPC_URL=http://localhost:8546
+WAVS_SERVICE_MANAGER_ADDRESS=`jq -r .addresses.WavsServiceManager .nodes/avs_deploy.json`
+MIRROR_SERVICE_MANAGER_ADDRESS=`jq -r .addresses.WavsServiceManager .nodes/mirror.json`
+
 docker run --rm --network host -v ./.nodes:/root/.nodes \
+   -e SOURCE_SERVICE_MANAGER_ADDRESS=${WAVS_SERVICE_MANAGER_ADDRESS} \
+   -e MIRROR_SERVICE_MANAGER_ADDRESS=${MIRROR_SERVICE_MANAGER_ADDRESS} \
+   -e SOURCE_RPC_URL=${SOURCE_RPC_URL} \
+   -e MIRROR_RPC_URL=${MIRROR_RPC_URL} \
   wavs-middleware -m mirror list_operators
```
Supeeerpower, line 153: All of these envs can be loaded from the .nodes directory.
If we can assume that DEV.md is for developers, you can skip this part.
It's not really important. It's just a suggestion.

PR #201, opened 2025-07-21
renaming refactor and avs-sync handler addition

noahsaso:
As @Reecepbcups, @ismellike, @Supeeerpower, and I discussed last week, there is confusion about where certain code and workflows belong. Our conclusion was:

- the sync services (`avs-sync`, `multi-chain-operator-sync`, `multi-chain-service-manager-sync`) should probably exist separate from `wavs-middleware`, which they do right now in `wavs-tools`, since they are not specific to the chain/VM being submitted to. `wavs-tools` may or may not be the right place for them, but it's fine for now.
- service handler contracts for the sync services *should* exist with the middleware, since they always have to be deployed for each service, and will vary based on the chain/VM (e.g. there will be different avs-sync contract handlers for EVM and CosmWasm, so the handlers should live with the middleware for the respective chain)
- the names need to be more descriptive

The main inconsistency was that `avs-sync`'s `AvsWriter.sol` service handler existed in `wavs-tools`, while `multi-chain-operator-sync` and `multi-chain-service-manager-sync`'s service handlers are in `wavs-middleware` and are automatically deployed with the middleware. To reconcile this, this PR moves `AvsWriter.sol` to this repo, and includes it in the middleware deploy automatically just like the rest of the sync service handlers.

The service handlers were also named confusingly, with `multi-chain-operator-sync`'s service handler called `MirrorServiceHandler.sol`, and `multi-chain-service-manager-sync`'s called `MirrorServiceManagerHandler.sol`. To reconcile this, this PR renames them to `MirrorOperatorSyncHandler.sol` and `MirrorQuorumSyncHandler.sol`, respectively. `avs-sync`'s `AvsWriter.sol` was moved to this repo and renamed to `WavsOperatorUpdateHandler.sol`.

The PR in `wavs-tools` (https://github.com/Lay3rLabs/wavs-tools/pull/96) renames `avs-sync` to `operator-updater` as well as `multi-chain-service-manager-sync` to `multi-chain-quorum-sync` to be more descriptive, which both reflect exactly what they do.

### DIFF
--- a/contracts/script/eigenlayer/ecdsa/WavsMiddlewareDeployer.s.sol
+++ b/contracts/script/eigenlayer/ecdsa/WavsMiddlewareDeployer.s.sol
@@ -47,6 +47,8 @@ contract WavsMiddlewareDeployer is Script, IECDSAStakeRegistryTypes {
     error WavsMiddlewareDeployer__DelegationManagerAddressCannotBeZero();
     /// @notice The error for the AVS directory address cannot be zero.
     error WavsMiddlewareDeployer__AVSDirectoryAddressCannotBeZero();
+    /// @notice The error for the operator update handler address cannot be zero.
+    error WavsMiddlewareDeployer__OperatorUpdateHandlerAddressCannotBeZero();
 
     /// @notice The setup function for the script.
     function setUp() public virtual {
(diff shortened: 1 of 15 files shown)


ismellike (approved):
lgtm

naming is more clear
AVSWriter got some nice improvements and some great coverage now too

Supeeerpower (approved):
Great, thank you @NoahSaso 
LGTM!

PR #202, opened 2025-07-22
200 Logo Upload

### LINKED ISSUE #200: Provide valid metadata URI default
We currently point to https://wavs.xyz/metadata.json in the env.example files.

There is nothing at that URI. We should create some default json file and add to this repo, and then upload that to some common location. That would be a [json file of this format](https://raw.githubusercontent.com/ethgas-developer/ethgas-developer.github.io/main/vision-avs-2.json) but with some references to "Made by WAVS" and "wavs-app" and linking our twitter. Which then could be customised by those who want to do that.

Supeeerpower:
Fixes on #200 

Uploaded Wavs logo file.


ethanfrey:
I think this was only part. Actually creating a metadata file and then using that URL as a default in the deploy scripts was the rest.

Or did that happen in another PR?

PR #207, opened 2025-07-23
205 Separate Script for Mock Deploy and Configure

### LINKED ISSUE #205: Document and maybe add CLI helpers for mock deployment
The [instructions currently say](https://github.com/Lay3rLabs/wavs-middleware?tab=readme-ov-file#1-create-a-configuration-file):

> Create a mock-config.json file on your local machine. This file defines the initial operators, their signing keys, weights, and the threshold for the stake registry.

This doesn't need to change, but when it's the _only_ way to add operators, it's problematic for WAVS which needs to know the service manager address _before_ it gets the AVS Key.

@ethanfrey recommended that we document an "upgrade" solution here, so that the contracts can be deployed with garbage/partial data, and then upgrade after deployment

Along with documentation, this upgrade should be doable by calling the docker image with a command, especially if this upgrade function is not on the IWavsServiceManager interface - that command should probably take:

1. serviceManagerAddress
2. operatorAddress
3. signingKey
4. weight

Supeeerpower:
Fixes on #205 

- Forge and shell script update to separate for deployment and configuration of mock contracts.
- README.md update.
- Added env reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,32 @@ docker run --rm --network host -v ./.nodes:/root/.nodes \
 
 This deployment process is for local testing and development. It deploys a "mock" version of the WAVS middleware contracts by using the mock stage of the mirror deployment scripts. This allows for rapid testing without needing to interact with a live EigenLayer environment.
 
-### 1. Create a Configuration File
+### 1. Run a Local Blockchain
+
+```bash
+anvil --host 0.0.0.0 --port 8546
+```
+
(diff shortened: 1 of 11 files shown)


dakom (approved):
I'm not so familiar with the internals of the shell scripts here, but the ability to deploy the service manager and then later configure it with operators is great!

ismellike:
Can confirm its working for me on wavs-tools

ismellike (approved):

PR #208, opened 2025-07-23
Unique RPC_URL for Local and Testnet

### LINKED ISSUE #206: Improve consistency on RPC_URL env vars
ref https://github.com/Lay3rLabs/wavs-tools/pull/97/commits/28d291693dc31c0bebed8095ce53bdd5f21d68eb

To call scripts on testnet, we have to provide the correct rpc env var name `LOCAL_ETHEREUM_RPC_URL` or `TESTNET_RPC_URL`, meaning we need to create workarounds over on wavs-tools' taskfile to make it happen.

We should just have 1 param name like the mirror scripts do for both local and testnet `SOURCE_RPC_URL` | `MIRROR_RPC_URL`, which in this case could be just `RPC_URL` or `EVM_RPC_URL` etc

Supeeerpower:
Fixes on #206 

- Made unique rpc_url env for local and testnet

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ Prepare the env file:
 ```bash
 CHAIN=
 cp docker/env.example.$CHAIN docker/.env
-# edit the RPC_URL for a paid testnet rpc endpoint, add funded key, and TESTNET_RPC_URL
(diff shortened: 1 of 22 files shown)


ismellike (approved):
lgtm thank you!

PR #212, opened 2025-07-24
204 Separate Script to Deposit and Register

### LINKED ISSUE #204: Separate Script to Deposit and Register
Because of configuration delay, deposit and register are not able to proceed in one forge script fully automatically.
We need to have two separated forge script to deposit and register, and make delay in shell script.

Supeeerpower:
Fixes on #204 

- Added separated script for deposit and register.
- Updated register shell script with delay of local and testnet mode.
- Updated some path-ignore in ci.
- Updated MD doc for CLI.

### DIFF
--- a/contracts/script/eigenlayer/bls/utils/WavsDepositIntoStrategyLib.sol
+++ b/contracts/script/eigenlayer/bls/utils/WavsDepositIntoStrategyLib.sol
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+import {console2} from "forge-std/console2.sol";
+import {IStrategyManager} from "@eigenlayer/contracts/interfaces/IStrategyManager.sol";
+import {IDelegationManager} from "@eigenlayer/contracts/interfaces/IDelegationManager.sol";
+import {IStrategy} from "@eigenlayer/contracts/interfaces/IStrategy.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+/**
+ * @title WavsDepositIntoStrategyLib
+ * @author Lay3rLabs
+ * @notice This library is used to deposit into a WAVS strategy.
+ * @dev This library is used to deposit into a WAVS strategy.
+ */
+library WavsDepositIntoStrategyLib {
+    using Strings for *;
+
+    /// @notice The error for the failed to approve LST tokens.
+    error WavsDepositIntoStrategy__FailedToApproveLSTTokens();
+    /// @notice The error for the failed to mint LST tokens.
+    error WavsDepositIntoStrategy__FailedToMintLSTTokens();
+
+    /**
+     * @notice The function to deposit into a strategy.
+     * @param _strategyManagerAddress The address of the strategy manager.
+     * @param _delegationManagerAddress The address of the delegation manager.
+     * @param _lstContractAddress The address of the LST contract.
+     * @param _lstStrategyAddress The address of the LST strategy.
+     * @param _operatorAddr The address of the operator.
+     * @param _stakeAmount The amount of LST to stake.
+     */
+    function depositIntoStrategy(
+        address _strategyManagerAddress,
+        address _delegationManagerAddress,
+        address _lstContractAddress,
+        address _lstStrategyAddress,
+        address _operatorAddr,
+        uint256 _stakeAmount
+    ) internal {
+        IStrategyManager strategyManager = IStrategyManager(_strategyManagerAddress);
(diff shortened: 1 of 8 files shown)


contracts/script/eigenlayer/bls/utils/WavsDepositIntoStrategyLib.sol
```diff
@@ -0,0 +1,119 @@
...
+     */
+    function depositIntoStrategy(
+        address _strategyManagerAddress,
+        address _delegationManagerAddress,
+        address _lstContractAddress,
+        address _lstStrategyAddress,
+        address _operatorAddr,
+        uint256 _stakeAmount
+    ) internal {
+        IStrategyManager strategyManager = IStrategyManager(_strategyManagerAddress);
```
ethanfrey, line 42: Okay this is the code that was pulled out from the register lib, right?

contracts/script/eigenlayer/bls/utils/WavsRegisterOperatorLib.sol
```diff
@@ -40,105 +36,9 @@ library WavsRegisterOperatorLib {
...
     Vm internal constant VM = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
 
-    /// @notice The error for the failed to mint LST tokens.
-    error WavsRegisterOperatorLib__FailedToMintLSTTokens();
-    /// @notice The error for the failed to approve LST tokens.
-    error WavsRegisterOperatorLib__FailedToApproveLSTTokens();
     /// @notice The error for the failed to get allocation configuration delay.
     error WavsRegisterOperatorLib__FailedToGetAllocationConfigurationDelay();
 
-    /**
```
ethanfrey, line 50: Taken from here?

contracts/script/eigenlayer/bls/utils/WavsRegisterOperatorLib.sol
```diff
@@ -160,14 +60,6 @@ library WavsRegisterOperatorLib {
         IAllocationManager allocationManager = IAllocationManager(allocationManagerAddress);
         OperatorSet memory opSetQuery = OperatorSet({avs: serviceManagerAddress, id: 0});
         if (!allocationManager.isMemberOfOperatorSet(operatorAddr, opSetQuery)) {
-            (bool success, bytes memory result) = address(allocationManager).staticcall(
```
ethanfrey, line 163: And the  ugly workaround gone 😄

scripts/bls/eigen/register.sh
```diff
@@ -38,6 +48,21 @@ ensure_balance "$OP_ADDR"
...
 echo "Delegate amount: $WAVS_DELEGATE_AMOUNT"
 
-# Register operator
 cd contracts || handle_error "Failed to change to contracts directory"
-forge script script/eigenlayer/bls/WavsRegisterOperator.s.sol -vvv --rpc-url "$RPC_URL" --private-key "$OPERATOR_KEY" --broadcast --skip-simulation --slow || handle_error "Failed to register operator"
+
+# Deposit into strategy
+forge script script/eigenlayer/bls/WavsDepositIntoStrategy.s.sol -vvv --rpc-url "$RPC_URL" --private-key "$OPERATOR_KEY" --broadcast || handle_error "Failed to deposit into strategy"
+
+# Wait based on environment
```
ethanfrey, line 56: Ah, very cool.

Run one script, wait the time in bash, then run the second part.
Good approach

scripts/bls/eigen/register.sh
```diff
@@ -38,6 +48,21 @@ ensure_balance "$OP_ADDR"
...
+if [[ "$DEPLOY_ENV" == "LOCAL" ]]; then
+    echo "Mining $ALLOCATION_DELAY blocks for local transaction confirmation..."
+    for ((i=1; i<=ALLOCATION_DELAY; i++)); do 
+        cast rpc evm_mine --rpc-url http://localhost:8545
+    done
+elif [[ "$DEPLOY_ENV" == "TESTNET" ]]; then
+    WAIT_TIME=$((12 * ALLOCATION_DELAY))
```
ethanfrey: This seems a bit arbitrary.

Could you do something like:

* calculate target block number (current + allocation delay)
* while current_block < target_block
  * sleep (5 * target - current)

This is more robust and doesn't have to wait 2x the expected time. You can change 5 with any other number you like, whatever would be a lower bound for block time (could even be 1, you just have a few extra queries, but that is cheap)

ethanfrey (approved):
The design looks good. Approve it.

The one request to improve would be to make the sleep a bit more robust as commented below. You could merge and do that in a future PR, or make that change, test it works, then merge. No need to block on my re-review for that change

contracts/script/eigenlayer/bls/utils/WavsDepositIntoStrategyLib.sol
Supeeerpower: Yes

